PR #65, opened 2024-01-28
A0-3158: Dockerize relayer

mikogs:
Two things will happen when push is made on the `master` branch:
* eth and zero contracts are deployed to bridgenet/devnet
* their artifacts and addresses are stored
* relayer is build and dockerized (using earlier addresses and artifacts)
* and it's deployed to devnet


krzysztofziobro (approved):

bartoszjedrzejewski (changes requested):
I have a small issue with placing the addresses file. Right now it looks like this:
most/addresses/bridgenet/dev/0196d55c6f49d4a489209de0c3218b36c3f349ef/bridgenet.json
I'd rather have only current state of the environment, so:
most/addresses/bridgenet/dev.json <- this is bridgenet.json file renamed.

bartoszjedrzejewski (approved):

PR #67, opened 2024-01-30
Deploy real oracle

fbielejec:


azero/scripts/deploy.ts
```diff
@@ -5,7 +5,7 @@ import GovernanceConstructors from "../types/constructors/governance";
 import Governance from "../types/contracts/governance";
 import Most from "../types/contracts/most";
 import Token from "../types/contracts/token";
-import TestOracleConstructors from "../types/constructors/test_oracle";
+import OracleConstructors from "../types/constructors/oracle";
```
obrok, line 8: Request: I think it would be a good idea to make the one deployed here conditional on env settings, i.e. deploy the test one locally and the real one on bridgenet. That way it will be easier to run a "reasonable" local deployment without adding in code from another repo.

obrok (changes requested):

azero/scripts/deploy.ts
fbielejec: What is the point of having the TestOracle anyway?
You don't want to bother running Feeder locally - fine, Most ignores stale entries and returns a default.
You want to test gas prices - you need the  proper setup with the real oracle and a feeder anyway

fbielejec: done

krzysztofziobro (approved):
looks fine

PR #68, opened 2024-02-03
A0-3960: Separate contract for paying out pocket money

krzysztofziobro:
Adds a separate contract that pays out "pocket money" for initial tx fees on Aleph Zero network.
- Separates "pocket money" pool from relayer fees pool
- Allows fee to be modifiable by an entity separate from the guardian committee: paying out AZERO for tx fees is not security critical


azero/contracts/money-box/contract/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+        pub fn owner(&self) -> AccountId {
+            self.owner
+        }
+
+        #[ink(message)]
+        pub fn set_owner(&mut self, new_owner: AccountId) -> Result<(), MoneyBoxError> {
+            self.ensure_admin()?;
+            self.owner = new_owner;
+            Ok(())
+        }
```
obrok, line 76: Suggestion: I'm a bit surprised by the naming here. The admin sets the owner, not the other way around?

krzysztofziobro: Hmm, right, naming is not the best I'll think for a better name instead of `owner` - this role is supposed to be able to make payouts to users, but not to make any administrative calls (most is an `owner` but not an `admin`).

fbielejec: treasurer (_skarbnik_)

azero/contracts/money-box/contract/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+        owner: AccountId,
+        admin: AccountId,
+    }
+
+    impl MoneyBoxContract {
+        #[ink(constructor)]
+        pub fn new(amount_to_pay: Balance) -> Self {
+            Self {
+                amount_to_pay,
+                owner: Self::env().caller(),
```
fbielejec, line 61: pass as an argument to make deploying easier

azero/contracts/most/lib.rs
```diff
@@ -436,12 +435,12 @@ pub mod most {
             Ok(self.data()?.request_nonce)
         }
 
-        /// Query pocket money
+        /// Query subsidy contract
         ///
-        /// An amount of the native token that is tranferred with every request
+        /// An address of the contract that may be used to subsidize bridging users
         #[ink(message)]
-        pub fn get_pocket_money(&self) -> Result<Balance, MostError> {
```
fbielejec, line 443: names are all over the place. Contract is called `money_box`. In the callee contract it's `subsidy_contract`. Constant is `MONEY_BOX_GAS_LIMIT`. Please make them uniform

azero/contracts/money-box/contract/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    impl MoneyBox for MoneyBoxContract {
+        #[ink(message)]
+        fn pay_out(&self, to: AccountId) {
+            if self.env().balance() < self.amount_to_pay {
+                self.env().emit_event(InsufficientFunds {
+                    current_balance: self.env().balance(),
+                });
+            } else if self.env().caller() == self.owner
```
fbielejec, line 128: ensure_owner ?

azero/contracts/money-box/contract/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+        admin: AccountId,
+    }
+
+    impl MoneyBoxContract {
+        #[ink(constructor)]
+        pub fn new(amount_to_pay: Balance) -> Self {
+            Self {
+                amount_to_pay,
+                owner: Self::env().caller(),
+                admin: Self::env().caller(),
```
fbielejec, line 62: instead of two fields and duplicated  ensure_a, ensure_b etc introduce an e.g. `Role` enum and a mapping AccountId -> Role. You can then have one common check function one query function and one update function

fbielejec (changes requested):
I don't like the naming and how it's inconsistent between contracts - personally I liked pocket_money and keeping with the lingo you could use PiggyBank (for contract) and e.g. Role::Treasurer for a role that makes payouts. YMMV

krzysztofziobro:
Closing as we can achieve similar behavior in Most and we prefer to pocket money amount change to be under Most's owner control instead of some external account

PR #71, opened 2024-02-05
advisory contracts

fbielejec:


azero/contracts/advisory/contract/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "advisory"
+version = "1.0.0"
+authors = ["Cardinal Cryptography"]
+edition = "2021"
+
+[dependencies]
+ink = { version = "=4.3.0", default-features = false }
+# shared = { path = "../../shared" }
```
obrok, line 9:
```suggestion
```

azero/contracts/advisory/contract/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+        #[ink(message)]
+        fn flip_emergency(&mut self) -> Result<(), AdvisoryError> {
+            self.ensure_owner()?;
+            self.emergency = !self.emergency;
+            self.env().emit_event(EmergencyChanged {
+                emergency: self.emergency,
+            });
+            Ok(())
+        }
+
```
obrok, line 39: Suggestion: perhaps it would be cleaner if this were a `set_emergency` instead? Given multiple advisories exist, it would be unfortunate if someone sets or unsets this by accident, for example, because they think they are interacting with another one?

eth/contracts/Advisory.sol
```diff
@@ -0,0 +1,41 @@
+// Most Advisory
```
obrok, line 1: Request: we don't need this on eth, remove.

azero/contracts/advisory/trait/lib.rs
```diff
@@ -0,0 +1,36 @@
+#![cfg_attr(not(feature = "std"), no_std, no_main)]
```
obrok, line 1: Question: why do we need this trait thing? Isn't this for interacting with the contract from another contract?

obrok (changes requested):

azero/contracts/advisory/trait/lib.rs
fbielejec: not strictly needed, maybe e2e test, but removing

eth/contracts/Advisory.sol
fbielejec: Done

azero/contracts/advisory/contract/lib.rs
fbielejec: Good idea, done

obrok (approved):

PR #75, opened 2024-02-06
Advised relayer

fbielejec:


relayer/scripts/entrypoint.sh
```diff
@@ -26,9 +26,12 @@ KEYSTORE_PATH=${KEYSTORE_PATH:-""}
 RELAYER_ID=${RELAYER_ID:-0}
 
 AZERO_MOST_METADATA=${AZERO_MOST_METADATA:-"/usr/local/most.json"}
+ADVISORY_METADATA=${ADVISORY_METADATA:-"/usr/local/advisory.json"}
 
 ARGS=(
   --name "guardian_${RELAYER_ID}"
+  --advisory-contract-address=$(get_address $AZERO_ADDRESSES_FILE advisory)
```
krzysztofziobro, line 33:
```suggestion
  --advisory-contract-addresses=$(get_address $AZERO_ADDRESSES_FILE advisory)
```
Same in `run.sh`

relayer/src/contracts/azero.rs
```diff
@@ -35,6 +35,31 @@ pub enum AzeroContractError {
     MissingOrInvalidField(String),
 }
 
+pub struct AdvisoryInstance {
+    pub contract: ContractInstance,
+    pub address: AccountId,
+    pub transcoder: ContractMessageTranscoder,
```
krzysztofziobro, line 41:
```suggestion
```

relayer/src/main.rs
```diff
@@ -94,26 +110,40 @@ async fn main() -> Result<()> {
...
     tasks.spawn(async move {
-        EthListener::run(config, azero_connection, eth_connection, redis_connection)
-            .await
-            .map_err(ListenerError::from)
+        EthListener::run(
+            config_rc2,
+            azero_signed_connection_rc1,
+            eth_connection_rc1,
+            redis_connection_rc1,
+            emergency_rc2,
```
krzysztofziobro, line 126: Using rc numbers starts to become confusing when we pass arcs with different counters to the same task.
Maybe we should suffix with the name of corresponding task and not just a number (or somehow make it uniform within one spawn, number in the opposite direction for example)

relayer/src/listeners/azero.rs
```diff
@@ -135,50 +143,60 @@ impl AlephZeroListener {
...
-                first_unprocessed_block_number, to_block
-            );
-
-            // Add the next block numbers now, so that there is always some block number in the set.
-            for block_number in first_unprocessed_block_number..=to_block {
-                add_to_pending(block_number + 1, pending_blocks.clone()).await;
-            }
+            match emergency.load(Ordering::Relaxed) {
+                true => {
+                    trace!("Advisory is in an emergency state");
```
krzysztofziobro, line 148: I think this should be higher than `trace`, maybe `warn`, this is potentially important info

relayer/src/listeners/eth.rs
```diff
@@ -91,49 +95,58 @@ impl EthListener {
...
-            );
+            match emergency.load(Ordering::Relaxed) {
+                true => {
+                    trace!("Advisory is in an emergency state");
+                    sleep(Duration::from_millis(999)).await;
+                    continue;
+                }
+                false => {
+                    // Query for the next unknowns finalized block number, if not present we wait for it.
+                    let next_finalized_block_number = get_next_finalized_block_number_eth(
```
krzysztofziobro, line 106: Typically, relayer will wait here for several minutes, so it's most likely that if the emergency flag is set, the relayer will still process next batch of blocks (with the current implementation).

Similar situation with azero listener

relayer/src/listeners/advisory.rs
```diff
@@ -0,0 +1,161 @@
...
+use super::AzeroListenerError;
+use crate::{
+    config::Config,
+    connections::azero::AzeroWsConnection,
+    contracts::{decode_bool_field, AdvisoryInstance, AzeroContractError},
+};
+
+pub struct AdvisoryListener;
+
+impl AdvisoryListener {
```
krzysztofziobro, line 30: My intuition would be that a better approach is to just query the state of the contract every 5 seconds or so.
Response time would be similar, but the code would be much simpler + less load on the node

relayer/src/listeners/eth.rs
```diff
@@ -91,49 +95,58 @@ impl EthListener {
...
-            );
-
-            log::info!(
-                "Processing events from blocks {} - {}",
-                first_unprocessed_block_number,
-                to_block
-            );
+            match emergency.load(Ordering::Relaxed) {
+                true => {
+                    trace!("Advisory is in an emergency state");
```
krzysztofziobro, line 100: same with trace here

krzysztofziobro (changes requested):

relayer/src/listeners/advisory.rs
krzysztofziobro: Alternatively, we could take advantage of the fact that we are already fetching events from all blocks, but not sure how complex sharing that with advisory listener would be

relayer/scripts/devnet-relayers-compose.yml
```diff
@@ -7,21 +7,21 @@ services:
       - DEV_MODE=true
     image: most-relayer
     network_mode: host
-    restart: on-failure
+    restart: on-failure:3
```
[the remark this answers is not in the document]
fbielejec: the effect of the previous flag would be that in case of a broken relayer it would be stuck in a constant restart loop, which is not immediately evident when listing docker processes and obfusates docker logs, that might aid in triangulating what went wrong.  This way the process is gone after trying 3 times and you can do a post-mortem easily.

relayer/src/contracts/azero.rs
fbielejec: something went wrong with this comment?

relayer/src/listeners/azero.rs
fbielejec: warn will just flood your logs

relayer/src/listeners/advisory.rs
fbielejec: yeah, good idea, I will think of sharing this via a channel or something similar

relayer/src/listeners/eth.rs
fbielejec: same reply :)

relayer/src/contracts/azero.rs
krzysztofziobro: Why do you need this fields in `AdvisoryInstance`? I don't see them used anywhere

relayer/src/listeners/azero.rs
krzysztofziobro: OK, that's a good idea, but nut sure if it won't get unnecessarily complex.
We could also sleep longer or make that duration configurable

relayer/src/main.rs
fbielejec: done

relayer/src/contracts/azero.rs
fbielejec: I don't need the transcoder now that the event subscription is gone, but the address is needed (for logging)

relayer/src/listeners/advisory.rs
```diff
@@ -0,0 +1,94 @@
...
+            }
+
+            if previous_emergency_state && !current_emergency_state {
+                info!("Previously set emergency state has been lifted");
+            }
+
+            emergency.store(current_emergency_state, Ordering::Relaxed);
+
+            // we sleep for about half a block production time before making another round of queries
+            thread::sleep(Duration::from_millis(500))
```
krzysztofziobro, line 91: Why sleep on the entire thread?

fbielejec: ide typo, fixed

krzysztofziobro (approved):

PR #76, opened 2024-02-07
E2E ETH -> AZERO check

maciejzelaszczyk:
Adds the actual check to the `ETH -> AZERO` transfer test. Reorganizes some functionality. Adds `cargo test` as the test harness. Adds simple docs.


e2e-tests/src/test/eth_to_azero.rs
```diff
@@ -0,0 +1,111 @@
...
+use ethers::{
+    core::types::Address,
+    signers::{coins_bip39::English, MnemonicBuilder, Signer},
+    utils,
+};
+use log::info;
+
+use crate::{azero, config::setup_test, eth};
+
+/// One-way `Ethereum` -> `Aleph Zero` transfer trough `most`.
```
krzysztofziobro, line 13:
```suggestion
/// One-way `Ethereum` -> `Aleph Zero` transfer through `most`.
```
?

krzysztofziobro (commented):
Please add e2e to formatting & linting in our Makefile
Additionally, I think it would make sense to have a make instruction that would run these tests in the recommended way

e2e-tests/src/test/eth_to_azero.rs
```diff
@@ -0,0 +1,111 @@
...
+
+    let balance_post_transfer: u128 = weth_azero_contract
+        .contract_read(
+            &azero_connection,
+            "PSP22::balance_of",
+            &[(*azero_account_keypair.account_id()).clone().to_string()],
+        )
+        .await?;
+    info!("balance post transfer: {:?}", balance_post_transfer);
+    assert_eq!(transfer_amount, balance_post_transfer.into());
```
krzysztofziobro, line 108: I would propose to check difference in balance before and after bridging, this way we might rerun the test without tearing down the network.

maciejzelaszczyk:
@krzysztofziobro I modified and added the `make` instructions.

krzysztofziobro (approved):

PR #77, opened 2024-02-08
A0 3706 signer azero

obrok:
This is part 1 of the work for the signer. The second PR will add eth support. The third one (if there is time) will move to proper tokio instead of threads and `spawn_blocking`.


krzysztofziobro (commented):
Looks solid.

One thing that I would like to see is some automated way of spawning full relayers + signers setup (or some instructions how to do this).

relayer/signer/src/main.rs
```diff
@@ -0,0 +1,196 @@
...
+};
+use vsock::{VsockAddr, VsockListener, VMADDR_CID_ANY};
+
+#[derive(Parser)]
+struct ServerArguments {
+    #[clap(short, long, default_value = "1234")]
+    port: u32,
+
+    #[clap(short, long)]
+    azero_key: Option<String>,
```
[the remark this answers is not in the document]
obrok: No, that's a leftover from a temporary design, I'll change it.

relayer/relayer/src/main.rs
```diff
@@ -58,18 +61,27 @@ async fn main() -> Result<()> {
...
+    let azero_ro_connection = Arc::new(azero::init(&config.azero_node_wss_url).await);
+
+    let azero_rw_connection = if let Some(cid) = config.signer_cid {
+        AzeroConnectionWithSigner::with_signer(
+            azero::init(&config.azero_node_wss_url).await,
+            cid,
+            config.signer_port,
+        )
+        .await
+        .expect("Cannot connect to signer")
```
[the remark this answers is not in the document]
obrok: 👍 main didn't use to return Result.

krzysztofziobro (approved):

relayer/relayer/src/connections/azero.rs
```diff
@@ -0,0 +1,179 @@
+use std::sync::{Arc, Mutex};
```
fbielejec, line 1: Why change the directory layout so much, `signer` could just live in a top level, just like the `relayer` did?
This will cause massive refactors for opened PRs

relayer/scripts/run.sh
```diff
@@ -18,19 +18,27 @@ function get_address {
 
 # --- ARGS
```
fbielejec, line 19: rename to `run_relayer` ?

relayer/relayer/src/connections/azero.rs
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    async fn send_tx_with_params<Call: TxPayload + Send + Sync>(
+        &self,
+        tx: Call,
+        params: ParamsBuilder,
+        status: TxStatus,
+    ) -> anyhow::Result<TxInfo> {
+        if let Some(details) = tx.validation_details() {
+            info!(target:"aleph-client", "Sending extrinsic {}.{} with params: {:?}", details.pallet_name, details.call_name, params);
```
fbielejec, line 129: target is confusing in this log, since it's clearly not coming from the aleph-client but from relayer?

relayer/relayer/src/connections/azero.rs
```diff
@@ -0,0 +1,179 @@
...
+            TxStatus::Finalized => progress.wait_for_finalized_success().await?.into(),
+            // In case of Submitted block hash does not mean anything
+            TxStatus::Submitted => {
+                return Ok(TxInfo {
+                    block_hash: Default::default(),
+                    tx_hash: progress.extrinsic_hash(),
+                })
+            }
+        };
+        info!(target: "aleph-client", "tx with hash {:?} included in block {:?}", info.tx_hash, info.block_hash);
```
fbielejec, line 163: same here, confusing target

relayer/relayer/src/connections/azero.rs
obrok: I wanted to make it a workspace so that they have common dependency versions.

relayer/relayer/src/connections/azero.rs
obrok: :+1:

PR #78, opened 2024-02-08
A0-3706: Two phase owner change

JanKuczma:
Introduces two-phase owner change mechanisms to the following contracts:
- azero:
   - Most
   - Token
   - Oracle
- eth
   - Most
   
Notes:
- `renounceOwnership()` has been disabled for eth Most contract:
https://github.com/Cardinal-Cryptography/most/blob/67783cef26d75da6f7c2cd193755f1709d01e944/eth/contracts/Most.sol#L79-L80


azero/contracts/shared/src/ownable2step.rs
```diff
@@ -0,0 +1,64 @@
+#![cfg_attr(not(feature = "std"), no_std, no_main)]
```
krzysztofziobro, line 1: Seems like this code isn't used currently, am I wrong?

azero/contracts/ownable/lib.rs
```diff
@@ -0,0 +1,129 @@
+#![cfg_attr(not(feature = "std"), no_std, no_main)]
```
krzysztofziobro, line 1: I think it might be good to somehow mark in the name that this is "ownable" with two-step owner change, and not just a "regular" one

azero/contracts/ownable/lib.rs
```diff
@@ -0,0 +1,129 @@
...
+
+#[derive(Debug, PartialEq, Eq, Encode, Decode)]
+#[cfg_attr(feature = "std", derive(scale_info::TypeInfo))]
+pub enum Error {
+    /// The caller didn't have the permissions to call a given method
+    UnauthorizedAccount(AccountId),
+    /// The caller tried to accept ownership but the process hasn't been started
+    NoPendingOwner,
+    /// Useful in cases, when the `Data` struct is not accessed directly but inside of `Lazy` or a `Mapping`, means that we failed to access the `Data` struct itself.
+    CorruptedStorage,
```
krzysztofziobro, line 27: I don't think it should be a part of this Error enum, as it's not an error originating from this struct, but from the contract that uses it

azero/contracts/ownable/lib.rs
```diff
@@ -0,0 +1,129 @@
...
+///
+/// In order to use it in your contract, implement the methods of the `Ownable2Step` trait: in most cases, you can simply call the corresponding methods on the `Data` object.
+use ink::primitives::AccountId;
+use scale::{Decode, Encode};
+
+#[derive(Debug, PartialEq, Eq, Encode, Decode)]
+#[cfg_attr(feature = "std", derive(scale_info::TypeInfo))]
+pub enum Error {
+    /// The caller didn't have the permissions to call a given method
+    UnauthorizedAccount(AccountId),
```
krzysztofziobro, line 23: Would change to `CallerNotOwner`, but just my preference, can be kept the way it is

azero/contracts/ownable/lib.rs
```diff
@@ -0,0 +1,129 @@
...
+/// * `accept_owership`: callable only by the pending owner, removes the previous owner and makes them the sole owner of the contract
+/// * `get_pending_owner`: returns the pending owner, if the ownership change process is currently underway.  
+///
+/// In order to use it in your contract, implement the methods of the `Ownable2Step` trait: in most cases, you can simply call the corresponding methods on the `Data` object.
+use ink::primitives::AccountId;
+use scale::{Decode, Encode};
+
+#[derive(Debug, PartialEq, Eq, Encode, Decode)]
+#[cfg_attr(feature = "std", derive(scale_info::TypeInfo))]
+pub enum Error {
```
krzysztofziobro, line 21: Would prefer `OwnableError` and `OwnableData`

azero/contracts/gas-price-oracle/contract/lib.rs
```diff
@@ -6,13 +6,15 @@ pub mod oracle {
...
     };
+    use ownable::*;
     use scale::{Decode, Encode};
 
     #[ink(storage)]
     pub struct Oracle {
-        /// oracle owner
-        owner: AccountId,
+        /// data for Ownable2Step - oracle owner
+        ownable_data: Lazy<ownable::Data, ManualKey<0xDEADBEEF>>,
```
krzysztofziobro, line 17: I don't think `ownable_data` needs to be `Lazy` here

krzysztofziobro (commented):
First batch of comments

obrok (changes requested):
I think we should also have it in governance (both on eth and azero side). That's in case we don't manage to integrate signet/gnosis safe.

azero/contracts/ownable/lib.rs
JanKuczma: I agree. It's been done this way because `From<MostError>` for `Ownable2Step::Error` cannot be implemented in the Most contract since it's in the external crate. I'll think about more sleek solution.

azero/contracts/shared/src/ownable2step.rs
JanKuczma: Right. It comes from the initial commit. It shall be removed before the merge.

azero/contracts/gas-price-oracle/contract/lib.rs
JanKuczma: Correct me if I'm wrong: Assuming that `OwnableData` struct won't change in any potential contract update that's right. We can get rid of Lazy here.

azero/contracts/ownable/lib.rs
JanKuczma: Anyway, implementing `From<MostError>` for `ownable2Step::Error` would cause recursive types (as `Form<ownable2Step::Error>` has been implemented for `MostError` too) 🤔.
- Should we abandon `OwnableResult<T>` including `ownable2Step::Error`, and return booleans instead of `Return<(),ownable2step::Error>` were appropriate?
Or
- Replace `ownable2step::Error::CorruptedStorage` with more generic `ownabl2step::Error::Custom(String)`... ?

@piotrMocz

piotrMocz: I'd be a fan of the second option, to stay within the `Result` state of mind ;) It's kinda annyoying and a bit backwards that the implementing contract influences the way we design the trait but first of all, we're writing this library specifically for this contract and secondly, this is mostly due to limitations of ink traits, so it's what we've got to work with

deuszx: Unfortunately, you won't escape it (i.e. that contract influences the `trait`). So I stopped treating contract's traits as traits in "normal programming" and treat them as "whole package" and extract the trait only to minimise the dependencies later on.

azero/contracts/ownable/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+
+[features]
+default = ["std"]
+std = [
+    "ink/std",
+    "scale/std",
+    "scale-info/std",
+]
+ink-as-dependency = []
+e2e-tests = []
```
krzysztofziobro, line 27: Are these needed here?

krzysztofziobro (approved):
Good

azero/contracts/ownable/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+/// * `accept_owership`: callable only by the pending owner, removes the previous owner and makes them the sole owner of the contract
+/// * `get_pending_owner`: returns the pending owner, if the ownership change process is currently underway.  
+///
+/// In order to use it in your contract, implement the methods of the `Ownable2Step` trait: in most cases, you can simply call the corresponding methods on the `Data` object.
+use ink::{prelude::string::String, primitives::AccountId};
+use scale::{Decode, Encode};
+
+#[derive(Debug, PartialEq, Eq, Encode, Decode)]
+#[cfg_attr(feature = "std", derive(scale_info::TypeInfo))]
+pub enum OwnableError {
```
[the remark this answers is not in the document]
fbielejec: nit: package could also be called ownable2step

azero/contracts/ownable/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+    NoPendingOwner,
+    /// Useful in cases, when the `Data` struct is not accessed directly but inside of `Lazy` or a `Mapping`, means that we failed to access the `Data` struct itself.
+    Custom(String),
+}
+
+pub type OwnableResult<T> = Result<T, OwnableError>;
+
+#[derive(Debug)]
+#[ink::storage_item]
+pub struct Ownable2StepData {
```
fbielejec, line 33: it's more idiomatic to call the struct without it's Data suffix, `data` is used for e.g. lazy storage items, to signify they are decoded all at once

azero/contracts/ownable/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+    /// The caller tried to accept ownership but the process hasn't been started
+    NoPendingOwner,
+    /// Useful in cases, when the `Data` struct is not accessed directly but inside of `Lazy` or a `Mapping`, means that we failed to access the `Data` struct itself.
+    Custom(String),
+}
+
+pub type OwnableResult<T> = Result<T, OwnableError>;
+
+#[derive(Debug)]
+#[ink::storage_item]
```
fbielejec, line 32: you don't need storage_item here since it is not an inner, non-packed storage type - the data goes under root key anyway - use  #[ink(storage)] instead.

azero/contracts/ownable/.gitignore
```diff
@@ -0,0 +1,9 @@
+# Ignore build artifacts from the local tests sub-crate.
+/target/
+
+# Ignore backup files creates by cargo fmt.
+**/*.rs.bk
```
obrok, line 5: Question: can't this be covered more generally in the overall `.gitignore` for the repo?

azero/contracts/ownable/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    pub fn transfer_ownership(
+        &mut self,
+        caller: AccountId,
+        new_owner: AccountId,
+    ) -> OwnableResult<()> {
+        if caller != self.owner {
+            return Err(OwnableError::CallerNotOwner(caller));
+        }
```
obrok, line 53:
```suggestion
        ensure_owner(caller)?;
```

no?

azero/contracts/ownable/lib.rs
```diff
@@ -0,0 +1,120 @@
...
+        self.pending_owner = Some(new_owner);
+
+        Ok(())
+    }
+
+    pub fn accept_ownership(&mut self, caller: AccountId) -> OwnableResult<()> {
+        let pending_owner = self.pending_owner.ok_or(OwnableError::NoPendingOwner)?;
+
+        if caller != pending_owner {
+            return Err(OwnableError::CallerNotOwner(caller));
```
obrok, line 64: Suggestion: make it a different error maybe? Something like `CallerNotPendingOwner`, which is more specific.

azero/scripts/deploy.ts
```diff
@@ -190,13 +195,72 @@ async function main(): Promise<void> {
...
+      [],
+      true,
+    );
+  //Members signing proposal...
+  console.log("Members signing proposal...");
+  await governance.withSigner(governance_members[1]).tx.vote(0);
+  await governance.withSigner(governance_members[2]).tx.vote(0);
+  //execute proposal
+  console.log("Executing proposal...");
+  await governance.tx.executeProposal(0);
```
obrok, line 218: Request: extract this as a function, please. It's gonna be a PITA to update otherwise.

eth/scripts/2_setup_contracts.js
```diff
@@ -45,6 +45,23 @@ async function main() {
...
+  let calldata = await iface.encodeFunctionData("acceptOwnership", []);
+  await governanceInstance
+    .connect(signers[1])
+    .submitProposal(contracts.governance, calldata);
+  console.log("Proposal submitted");
+  console.log("Signing proposal...");
+  await governanceInstance.connect(signers[2]).vote(0);
+  await governanceInstance.connect(signers[3]).vote(0);
+  // execute proposal
+  await governance.executeProposal(0);
```
obrok, line 64: Suggestion: maybe something could be extracted as a function here as well?

obrok (changes requested):

PR #79, opened 2024-02-10
A0-4045: Fix the issue with pocket money being paid out from rewards

krzysztofziobro:
Fixes the issue in which "pocket money" might have been paid out from funds assigned to the guardian committee


azero/contracts/most/lib.rs
```diff
@@ -276,10 +279,18 @@ pub mod most {
...
+            if let Some(surplus) = transferred_fee.checked_sub(current_base_fee) {
                 self.env().transfer(sender, surplus)?;
             };
 
+            self.env().emit_event(CrosschainTransferRequest {
+                committee_id: data.committee_id,
+                dest_token_address,
+                amount,
+                dest_receiver_address,
+                request_nonce: data.request_nonce.saturating_sub(1),
```
fbielejec, line 291: why subtracting 1 here?

krzysztofziobro: changed the ordering so that emission is at the end - request_nonce has already been updated

azero/contracts/most/lib.rs
```diff
@@ -350,11 +364,16 @@ pub mod most {
...
                 // bootstrap account with pocket money
-                // NOTE: we don't revert on a failure!
-                _ = self
-                    .env()
-                    .transfer(dest_receiver_address.into(), data.pocket_money);
+                if self.env().balance() >= unclaimed_rewards_all.saturating_add(data.pocket_money) {
+                    // don't revert if the transfer fails
+                    _ = self
+                        .env()
+                        .transfer(dest_receiver_address.into(), data.pocket_money);
```
obrok, line 375: Question: Do I understand this correctly?

1. We pay out the pocket money only if the balance of the contract is more than the rewards. 
2. So in effect we prioritize paying out the rewards over paying out pocket money.
3. The money for those two activities is mixed in the contract balance.

I'm a bit worried that this setup will make it difficult to manage the balance available for pocket money. In particular, if we have a script that says "make sure this account has at least $1000", could it be that someone does not collect their rewards and, after some time, we stop reloading the account, because the balance is too large, but all of that money is "claimed" by guardians?

krzysztofziobro: Yes, that's correct. 
I might add a helper method for fetching balance that can be distributed as pocket money, do you think it would be useful?

azero/contracts/most/lib.rs
fbielejec: then I suggest:
- get the nonce from storage (it has copy semantics)
- update it in the storage, use checked math
- use original value in the event emission.

cheaper in gas and less confusing to read

azero/contracts/most/lib.rs
fbielejec: alternatively we make topping up via a tx and keep track of the balance for pocket monies in it' sown storage?

krzysztofziobro: With the current fee structure it would be assumed to be neither rewards nor balance for pocket money (as we track rewards explicitly)

fbielejec: yes, so you'd track rewards and pocket money explicitely, every time you send pocket money you decrease it, every time it is topped up (with a tx) you increase it. Only send out if you leave positive balance afterwards

The downside is that you cannot send A0 with a normal balances transfer this way, you need to do it with a contract tx, or else the moneyz is stuck

krzysztofziobro: So you track both rewards and pocket money? What advantage does it have over the current approach?

obrok: Well, I'm thinking of the person who has to do the top-up - it's either going to be a script, or someone is going to do this manually. If script, then I assume (because the script doesn't exist yet) that it will just look at the account balance (because we won't remember/the person writing the script won't know about this corner case), which will cause this problem. If manually, then the person doing the top-up will need some way to check, so perhaps a contract method that exposes remaining pool of pocket money would be useful?

fbielejec: I guess simple, explicit & easier to read - also easy to see when exactly to top up - one variable in storage vs having two

krzysztofziobro: OK, will rewrite it that way

fbielejec: your call - I'm just suggesting it as an option (not a blocking request) . The fact you have to top up with a transaction can be seen as a potential downside

azero/contracts/most/lib.rs
```diff
@@ -427,6 +437,19 @@ pub mod most {
...
+        /// Method used to provide the contract with funds for pocket money
+        #[ink(message, payable)]
+        pub fn fund_pocket_money(&mut self) -> Result<(), MostError> {
+            let funds = self.env().transferred_value();
+            let mut data = self.data()?;
+            data.pocket_money_balance = data
+                .pocket_money_balance
+                .checked_add(funds)
+                .ok_or(MostError::Arithmetic)?;
+            self.data.set(&data);
```
fbielejec, line 449: maybe emit an event?

fbielejec (approved):

azero/contracts/most/lib.rs
```diff
@@ -377,10 +381,16 @@ pub mod most {
...
-                    .transfer(dest_receiver_address.into(), data.pocket_money);
+                if data.pocket_money_balance >= data.pocket_money {
+                    // don't revert if the transfer fails
+                    _ = self
+                        .env()
+                        .transfer(dest_receiver_address.into(), data.pocket_money);
+                    data.pocket_money_balance = data
+                        .pocket_money_balance
+                        .checked_sub(data.pocket_money)
+                        .ok_or(MostError::Arithmetic)?;
```
obrok, line 419: Question: I think this is inconsistent, no? We don't revert if the transfer fails, but we revert if this subtraction fails? I guess the point is that if this fails, then we have a bug in the math somewhere? Overall, now that I'm seeing this solution, I'm somewhat concerned - it introduces these two pieces of state that must be kept consistent with each other, but there is nothing that guarantees it - for example, in case of some refactoring. Perhaps we should have a struct encapsulating the two fields that makes sure we always update them correctly if you want to keep track of both values? I'm genuinely unsure of what's the best course of action here.

obrok (approved):
Seems fine, although I have those concerns I mentioned.

azero/contracts/most/lib.rs
krzysztofziobro: Yes, failure here signals a bug in the contract, and for the transfer to fail there would also need to be a bug, so I think I get your point about inconsistency. Do you think both these should revert on failure?

Can you name these two pieces of state? Do you mean the part that keeps track of the rewards and the part that keeps track of pocket money balance? They are both updated along with value transfer/receive which I think is strong enough (unless the current code has some bug that I don't see). I am not sure if I get your point here

PR #80, opened 2024-02-11
A0-3158: Add if-exist input to steps in s3 contracts

mikogs:
Add `if-exist` input with value of `overwrite` to overwrite files in S3 bucket if they exist.

### DIFF
--- a/.github/workflows/contracts-compile-and-deploy-to-bridgenet.yml
+++ b/.github/workflows/contracts-compile-and-deploy-to-bridgenet.yml
@@ -103,7 +103,7 @@ jobs:
           path: eth
 
       - name: Copy files to S3 AWS bucket
-        uses: Cardinal-Cryptography/github-actions/copy-file-to-s3@v1
+        uses: Cardinal-Cryptography/github-actions/copy-file-to-s3@v6
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/contracts-compile-and-deploy-to-bridgenet.yml
```diff
@@ -103,7 +103,7 @@ jobs:
           path: eth
 
       - name: Copy files to S3 AWS bucket
-        uses: Cardinal-Cryptography/github-actions/copy-file-to-s3@v1
+        uses: Cardinal-Cryptography/github-actions/copy-file-to-s3@v6
```
Marcin-Radecki, line 106: Even though there's a default value to newly added `if-exists` argument in this action, this default value changes behavior (logc) if the action - now we skip when file exists, and previously we did Overwrite. Then I think we need to bump it to `v7`.

Marcin-Radecki (approved):

.github/workflows/contracts-compile-and-deploy-to-bridgenet.yml
Marcin-Radecki: I was mistaken, action does not change its behaviour.

PR #83, opened 2024-02-13
Propagate revert error to Azero Listener + adding nonce manager for eth transactions

krzysztofziobro:
- Adding nonce management for eth txs: now these transactions will be handled properly in parallel eg. when recovering from a crash
- Propagate info about revert into Azero Listener - resulting in a crash
- Changed contracts so they do not revert when relayer votes for the same transfer for the second time


relayer/src/connections/eth.rs
```diff
@@ -30,5 +32,8 @@ pub async fn sign(
     connection: EthConnection,
     wallet: LocalWallet,
 ) -> Result<SignedEthConnection, EthConnectionError> {
-    Ok(SignerMiddleware::new_with_provider_chain(connection, wallet).await?)
+    let nonce_manager = connection.nonce_manager(wallet.address());
```
fbielejec, line 35: nice!

fbielejec (approved):
rebase & LGTM

PR #88, opened 2024-02-16
Fix bridgenet deployment

fbielejec:
- in stupidass hardhat `address` is actually `target`
- hotfix for gov addresses without funds

### DIFF
--- a/.github/workflows/on-master-branch-commit.yml
+++ b/.github/workflows/on-master-branch-commit.yml
@@ -5,7 +5,6 @@ on:
   push:
     branches:
       - master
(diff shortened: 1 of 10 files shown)


.github/workflows/on-master-branch-commit.yml
```diff
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - master
+      - fix-bridgenet-deployment
```
obrok, line 8:
```suggestion
```

obrok (approved):

JanKuczma (approved):

PR #89, opened 2024-02-16
setup script hangs forever

fbielejec:


eth/scripts/2_setup_contracts.js
```diff
@@ -114,6 +114,8 @@ async function main() {
   await migrations.setCompleted(2);
 
   console.log("Done");
+  // NOTE: neccessary because script hangs in CI
+  process.exit(0);
```
obrok, line 118: O.o

fbielejec: if it's stupid but works it's not stupid :man_shrugging:

krzysztofziobro (approved):

PR #91, opened 2024-02-16
A0-3706 ETH transactions in signer

obrok:
Adds support for ETH transactions in signer. I'll make another PR where I dockerize it for infra usage.

### DIFF
--- a/relayer/relayer/src/connections/eth.rs
+++ b/relayer/relayer/src/connections/eth.rs
@@ -1,15 +1,25 @@
+use std::sync::{Arc, Mutex};
+
 use ethers::{
+    abi::Address,
+    core::{types::H256, utils::hash_message},
     prelude::{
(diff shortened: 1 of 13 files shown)


relayer/relayer/src/connections/eth.rs
```diff
@@ -1,39 +1,235 @@
...
+    let client = signer_client::Client::new(cid, port)?;
+    let client = Arc::new(Mutex::new(client));
+    let client_rc = client.clone();
+    let address = spawn_blocking(move || {
+        let client = client_rc.lock().unwrap();
+        client.eth_address()
+    })
+    .await??;
+
+    let nonce_manager = connection.nonce_manager(address);
```
krzysztofziobro, line 225: I think you forgot to initialize the nonce manager here

obrok: :+1:

PR #93, opened 2024-02-19
A0-4045 bridging native eth

JanKuczma:
Allow users to implicitly bridge native etherum and receive native etherum when bridging back.

- adds `sendRequestNative(...)` to Most.sol contract
- changes `receiveRequest(...)` in Most.sol to send native ETH when weth is being bridged
- unifies naming of WETH9 instances to `weth`


eth/contracts/Most.sol
```diff
@@ -115,6 +117,35 @@ contract Most is Initializable, UUPSUpgradeable, Ownable2StepUpgradeable {
...
+    function sendRequestNative(
+        bytes32 destReceiverAddress
+    ) external payable {
+        uint256 amount = msg.value;
+
+        // check if the token is supported
+        bytes32 destTokenAddress = supportedPairs[addressToBytes32(wethAddress)];
+        require(destTokenAddress != 0x0, "Unsupported pair");
+
+        (bool success, ) = wethAddress.call{value: amount}(abi.encodeWithSignature("deposit()"));
```
krzysztofziobro, line 135: Since `wethAddress` is `payable` you should be able to use `weth.send` here. (calling `deposit` directly should not be required, but you can check)

eth/contracts/Most.sol
```diff
@@ -159,12 +190,19 @@ contract Most is Initializable, UUPSUpgradeable, Ownable2StepUpgradeable {
...
-            IERC20 token = IERC20(bytes32ToAddress(destTokenAddress));
-
-            token.transfer(
-                bytes32ToAddress(destReceiverAddress),
-                amount
-            );
+            if (bytes32ToAddress(destTokenAddress) == wethAddress ) {
+                (bool unwrapSuccess, ) = (wethAddress).call(abi.encodeWithSignature("withdraw(uint256)", amount));
+                require(unwrapSuccess, "Failed to uwrap wrapped ETH to native ETH.");
+                (bool sendNativeEthSuccess, ) = bytes32ToAddress(destReceiverAddress).call{value: amount}("");
```
krzysztofziobro, line 196: Again, use `send` if possible

JanKuczma (commented):
Actually, there's one case when the Most contract receives native eth other than through "sendRequestNative": when user is bridging the eth back, the Most contract unwraps weth and receives native eth (from the weth contract) which is then sent to the user. Hence, we need the receive in the Most contract.

krzysztofziobro:
Yes, that is a good point, my bad. Also, I see that actually `send` isn't a recommended way of making transfers, resolving my ignorant comments

krzysztofziobro (approved):
Good job!

PR #98, opened 2024-02-21
E2e azero -> eth

maciejzelaszczyk:
This is the leg of the test verifying that a return `Aleph Zero -> Ethereum` transfer works as expected.


e2e-tests/src/test/azero_to_eth.rs
```diff
@@ -0,0 +1,109 @@
...
+    let approve_info = weth_azero
+        .contract_exec(&azero_signed_connection, "PSP22::approve", &approve_args)
+        .await?;
+    info!("`approve` tx info: {:?}", approve_info);
+
+    loop {
+        if azero_signed_connection.get_finalized_block_hash().await? == approve_info.block_hash {
+            break;
+        }
+    }
```
krzysztofziobro, line 57: This seems to be unnecessary + looping this way looks suspicious

krzysztofziobro (approved):
Just one weird-looking loop, otherwise good

e2e-tests/src/test/azero_to_eth.rs
maciejzelaszczyk: OK, I have verified that we do not need that.

PR #99, opened 2024-02-22
A0-3990: Add a Makefile target to print contract hashes

obrok:
This includes modifying the dockerfile for the azero builder to be based on rust:1.70 instead of docker-ink-dev. Since rust:1.70 is an image maintained by an independent org on dockerhub it's a way to show "nothing up our sleeve" with this compilation.

### DIFF
--- a/docker/most_builder.dockerfile
+++ b/docker/most_builder.dockerfile
@@ -1,4 +1,4 @@
-FROM public.ecr.aws/p6e8q1z1/ink-dev:1.7.0
+FROM rust:1.70
 
 RUN apt update
 RUN apt install -y curl make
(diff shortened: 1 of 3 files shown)


docker/most_builder.dockerfile
```diff
@@ -11,6 +11,5 @@ RUN . "$NVM_DIR/nvm.sh" && nvm use v${NODE_VERSION}
 RUN . "$NVM_DIR/nvm.sh" && nvm alias default v${NODE_VERSION}
 ENV PATH="/root/.nvm/versions/node/v${NODE_VERSION}/bin/:${PATH}"
 
-COPY * .
-# Trigger toolchain install
-RUN cargo check || true
+COPY rust-toolchain.toml .
```
deuszx, line 14: Since you're setting the `--toolchain` manually in `typechain compile` (in `package.json`) this won't be used for contracts. It will be for relayer though.

deuszx (approved):
Q: what's the size diff between the new docker and previous one?

Also, remember that reusing the same Dockerfile may lead to the deterministic build and you'll need a pre-built docker image for that.

krzysztofziobro (approved):

PR #100, opened 2024-02-22
Error handling improvements

krzysztofziobro:
In this PR:
- automated restart of AzeroListener in case of pause + unpause
- different way of overriding redis cache that allows for previous
- added sleep to the advisory handling: previous version would needlessly spam node with requests
- removed special treatment of CommitteeId mismatch

### DIFF
--- a/relayer/relayer/src/main.rs
+++ b/relayer/relayer/src/main.rs
@@ -153,15 +153,10 @@ async fn main() -> Result<()> {
     )
     .await
     {
-        error!("Error when running listeners, this might require manual investigation...");
-        err.chain().enumerate().for_each(|(level, cause)| {
-            let cause = cause.to_string();
-            if cause.len() > 100 {
-                error!(" {}: {}...", level, &cause[..100]);
-            } else {
-                error!(" {}: {}", level, cause);
-            }
-        });
+        error!(
+            "Error when running listeners, this might require manual investigation...: {:?}",
+            err
+        );
     }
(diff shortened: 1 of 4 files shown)


relayer/relayer/src/main.rs
```diff
@@ -139,47 +144,148 @@ async fn main() -> Result<()> {
...
+        // For most errors we don't want to automatically restart the relayer but wait for manual intervention
+        loop {
+            error!("Error when running listeners, this might require manual investigation or RESTART...");
+            err.chain().enumerate().for_each(|(level, cause)| {
+                let cause = cause.to_string();
+                if cause.len() > 100 {
+                    error!(" {}: {}...", level, &cause[..100]);
+                } else {
+                    error!(" {}: {}", level, cause);
+                }
```
obrok, line 166: Suggestion: print the whole thing. Why are you worried about `cause` being longer than 100?

obrok (approved):
One suggestion, but may be irrelevant if you just Let It Crash™ as discussed.

relayer/relayer/src/main.rs
obrok: Hm, unfortunate, but maybe we can only shorten that particular one? Better yet - if we wrap it in some error of our own, implement the shortening in there.

krzysztofziobro: Changed it to print the whole error once

PR #102, opened 2024-02-24
Basic drink! tests.

deuszx:


azero/package.json
```diff
@@ -7,6 +7,8 @@
     "scripts": {
         "watch": "gulp default",
         "compile": "typechain-compiler --release --toolchain nightly-2024-02-16",
+        "predrink-tests": "npm config set script-shell \"/bin/bash\"; ./scripts/predrink_tests.sh",
```
deuszx, line 10: I probably shouldn't run it on every `npm run drink-tests` as it'll rerun ink-wrapper binary which is not always necessary.

TODO: Make it into a standalone command for generating wrappers.

PR #103, opened 2024-02-26
ink Most contract: expose vote/signature related getters

krzysztofziobro:
Adds getters for:
- request signature/vote count
- whether a given request has been executed
- whether a specified guardian has signed some specific request

Useful for displaying transactions status in front-end.

### DIFF
--- a/azero/contracts/most/lib.rs
+++ b/azero/contracts/most/lib.rs
@@ -747,6 +747,29 @@ pub mod most {
             Ok(self.data()?.is_halted)
         }
 
+        /// Returns the signature count for a given unprocessed request.
+        /// Will return 0 if the request has already been processed.
+        #[ink(message)]
+        pub fn signature_count(&self, hashed_request: HashedRequest) -> u128 {
+            if let Some(Request { signature_count }) = self.pending_requests.get(hashed_request) {
+                signature_count
+            } else {
+                0
+            }
+        }
+
+        /// Returns whether the given account has signed the given request
+        #[ink(message)]
+        pub fn has_signed(&self, hashed_request: HashedRequest, account: AccountId) -> bool {
+            self.signatures.contains((hashed_request, account))
+        }
+
+        /// Returns whether the given transfer request has been executed
+        #[ink(message)]
+        pub fn has_been_executed(&self, hashed_request: HashedRequest) -> bool {
+            self.processed_requests.contains(hashed_request)
+        }
+
         // ---  helper functions
 
         fn check_halted(&self) -> Result<(), MostError> {


azero/contracts/most/lib.rs
```diff
@@ -747,6 +747,29 @@ pub mod most {
...
+                signature_count
+            } else {
+                0
+            }
+        }
+
+        /// Returns whether the given account has signed the given request
+        #[ink(message)]
+        pub fn has_signed(&self, hashed_request: HashedRequest, account: AccountId) -> bool {
+            self.signatures.contains((hashed_request, account))
```
obrok, line 764: Suggestion: not sure about this one, but could also return a separate value if the hash is invalid.

azero/contracts/most/lib.rs
```diff
@@ -747,6 +747,29 @@ pub mod most {
             Ok(self.data()?.is_halted)
         }
 
+        /// Returns the signature count for a given unprocessed request.
+        /// Will return 0 if the request has already been processed.
```
obrok, line 751: Suggestion: make it return something like:

```rust
enum SignatureCount {
  Pending { needed: u128 },
  AlreadyProcessed,
  InvalidHash
}
```

azero/contracts/most/lib.rs
```diff
@@ -747,6 +747,29 @@ pub mod most {
...
+        /// Returns whether the given account has signed the given request
+        #[ink(message)]
+        pub fn has_signed(&self, hashed_request: HashedRequest, account: AccountId) -> bool {
+            self.signatures.contains((hashed_request, account))
+        }
+
+        /// Returns whether the given transfer request has been executed
+        #[ink(message)]
+        pub fn has_been_executed(&self, hashed_request: HashedRequest) -> bool {
+            self.processed_requests.contains(hashed_request)
```
obrok, line 770: Suggestion: same here, could return an error for invalid hashes.

krzysztofziobro: By invalid you mean a hash that's not processed and no votes have been cast for its execution?

azero/contracts/most/lib.rs
```diff
@@ -747,6 +755,22 @@ pub mod most {
             Ok(self.data()?.is_halted)
         }
 
+        /// Returns the status of a given cross-chain transfer request
+        #[ink(message)]
+        pub fn request_status(&self, hashed_request: HashedRequest) -> RequestStatus {
```
obrok, line 760: Question: I like this version much better, but you also removed the `has_signed` method - was that not needed for anything?

obrok (approved):

fbielejec (approved):

azero/contracts/most/lib.rs
krzysztofziobro: It could be used eg by relayer, but I don't think it is useful in the front-end

PR #107, opened 2024-02-27
Gnosis safe

fbielejec:
- creates a Safe instance in development and bridgenet networks
- signs & executes addPair tx with a Safe governance action


eth/scripts/0_deploy_gnosis.js
```diff
@@ -0,0 +1,145 @@
+const fs = require("node:fs");
+const Safe = require("@safe-global/protocol-kit").default;
+const { SafeFactory, EthersAdapter } = require("@safe-global/protocol-kit");
+const { network, ethers } = require("hardhat");
+
+async function main() {
+  switch (network.name) {
+    case "development" || "hardhat":
```
JanKuczma, line 8: Is there a reason why we don't want this to run on the bridgenet? Otherwise, lgtm

fbielejec: ooops typo - obviously we want it on bridgenet, fixed

JanKuczma (approved):

PR #110, opened 2024-02-29
create a safe on sepolia

fbielejec:


obrok (commented):
Question: why do we need this? Sepolia works from the official UI.

fbielejec:
> Question: why do we need this? Sepolia works from the official UI.

1. Because scripted, reproducable & automated > manual. 
2. All other deployment steps are scripted and under version control.
3. Because we will have to manage multiple environments and will quickly loose control and overview is we mix manual and scripted/automated deployments

JanKuczma (approved):

PR #114, opened 2024-03-01
Review eth contract

kroist:


eth/contracts/Most.sol
```diff
@@ -257,6 +264,9 @@ contract Most is
         return bytes32(uint256(uint160(addr)));
     }
 
+    // Some event like CommitteeUpdate might be Emitted here and in constructor
+    // For committee members to track their participation in comittee
+
```
fbielejec, line 272: added - but hwy emit in the constructor ? it always starts with an id of 0 and there is nothing to listen to before the contract is initialized ?

eth/contracts/Most.sol
```diff
@@ -100,6 +100,12 @@ contract Most is
     //
     // Tx emits a CrosschainTransferRequest event that the relayers listen to
     // & forward to the destination chain.
+
+    // let's use modern standards.
```
fbielejec, line 107: can you clarify this comment a bit, I'm not sure I follow.

Regardless of whether the allowance is increased via `approve` or `permit` the end result is the same - Most can spend up to a `value` on behalf of the sender, so it's more of a concern for the fronted to distinguish if token implement the ERC and whether to ask the user to sign a `permit` or just a plain `approval` tx ?

kroist: `permit` allows someone to execute `approve` on behalf of someone else. For the bridge frontend client, it might not be that useful. But for some users of bridge, who want to cross-contract call the bridge it might be very useful.

eth/contracts/Most.sol
kroist: agree, constructor emit might be ommited.

PR #115, opened 2024-03-01
Relayer suggestions

ggawryal:


relayer/signer/src/main.rs
```diff
@@ -162,6 +162,14 @@ async fn do_handle_client(
...
 
             Command::SignEthTx { mut tx, chain_id } => {
+                // [Audit] Have we considered checking somehow type of the signed transaction,
+                // or even changing the interface to accept only transactions that are calls
+                // to Most or Advisory contracts? Don't know how enclaves exactly work, but I
+                // assume that if an attacker can broke into the EC2 instance, then they
+                // could create extra connection to the enclave on which the signer is running,
+                // and sing anything they want, including draining all AZero and ETH from the guardian.
+                // This already isn't that bad, compared to having the key compromised, as at least we'd
+                // have some tracks of the incident probably, but maybe we could add that extra layer of security?
```
obrok, line 172: I specifically didn't want to do that to keep the signer as simple as possible to limit other possible bugs. So my choice was to have it be an abstract signer of bytes. From what I understand it would also be quite unwieldy to deploy - any change to the signer (not the relayer) requires rebuilding the enclave, giving it access to the signing keys, etc. - relatively high-security operations. So if anything simple happened where we needed to sign a new kind of transaction (like in fact we're planning to do when we finally add the ability for the relayer to pay out fees) we would have to go through that painful process.

PR #122, opened 2024-03-04
[Do not merge] Audit review

DamianStraszak:


azero/contracts/most/lib.rs
```diff
@@ -25,6 +25,7 @@ pub mod most {
     use shared::{concat_u8_arrays, keccak256, Keccak256HashOutput as HashedRequest, Selector};
 
     type CommitteeId = u128;
+    // We can save some gas by using a smaller type here -- u32 is enough.
```
DamianStraszak, line 28: Although I can now see that changing this now is a disaster. So Treat all the suggestions to change types with a grain of salt...

PR #125, opened 2024-03-05
Prepare deployments scripts for testnet deployment

fbielejec:
- configurable Oracle owner account
- explicit owners in config
- Migrations on Azero
- 0 deploys Migrations, every script updates it
- checks if previous migrations has been completed


azero/contracts/migrations/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+        #[ink(constructor)]
+        pub fn new(owner: AccountId) -> Self {
+            Self {
+                last_completed_migration: 0,
+                owner,
+            }
+        }
+
+        #[ink(message)]
+        pub fn set_completed(&mut self, completed: u32) {
```
JanKuczma, line 40: I think there is an "is_owner" check missing

azero/contracts/migrations/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+edition = "2021"
+
+[dependencies]
+ink = { version = "=4.3.0", default-features = false }
+
+scale = { package = "parity-scale-codec", version = "3", default-features = false, features = ["derive"] }
+scale-info = { version = "2.6", default-features = false, features = ["derive"], optional = true }
+
+[dev-dependencies]
+ink_e2e = "4.2.0"
```
JanKuczma, line 14:
```suggestion
```
No need for it, right?

azero/contracts/migrations/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
+name = "migrations"
+version = "0.1.0"
+authors = ["[your_name] <[your_email]>"]
```
JanKuczma, line 4:
```suggestion
authors = ["Cardinal Cryptography"]
```

JanKuczma (changes requested):

azero/contracts/migrations/Cargo.toml
fbielejec: indeed, removed

azero/contracts/migrations/lib.rs
fbielejec: oh yeah, very good catch!

PR #131, opened 2024-03-07
Refactor relayer, add circuit breaker

fbielejec:


relayer/relayer/src/listeners/azero.rs
```diff
@@ -1,238 +1,158 @@
...
+                                .send(AzeroMostEvents {
+                                    events: filtered_events.clone (),
+                                    events_ack_sender,
+                                }) => {
+                                    info!(target: "AlephZeroListener", "Sending a batch of {} events", &filtered_events.len());
+                                },
                         }
+
+                        info!(target: "AlephZeroListener", "Awaiting events ack");
+                        events_ack_receiver.await.map_err(|_| AlephZeroListenerError::AckSenderDropped)?;
```
krzysztofziobro, line 150: Question: this blocks until all transactions from that batch are finalized, right?
That would mean that bridge does not listen on Aleph Zero and submit eth transcactions until the whole previous batch is finalized. This isn't ideal, as it doubles (or more) bridging duration of many transactions

relayer/relayer/src/handlers/azero.rs
```diff
@@ -0,0 +1,267 @@
...
+                        .send(())
+                        .map_err(|_| AlephZeroEventsHandlerError::EventsAckReceiverDropped)?;
+                    info!("All events acknowledged");
+                }
+            }
+        }
+    }
+}
+
+async fn wait_for_eth_tx_finality(
```
krzysztofziobro, line 241: As discussed, it would be better to add a getter (in the contract) for checking if this guardian needs to sign the request and using https://docs.rs/ethers/latest/ethers/middleware/contract/struct.FunctionCall.html#method.block (in a loop):
1. check in the latest block if guardian's vote is needed
- if no: (1) wait for the time usually needed for finalization (2) check if now it is true also for the most recent finalized block
- if the vote is needed: submit it, wait the time needed, check if it's still needed etc.

krzysztofziobro: Similar for submitting to Aleph, but there we can skip the waiting part and simply submit, but with first checking that submitting the tx is needed

azero/scripts/bridge_weth.ts
```diff
@@ -0,0 +1,67 @@
+import { ApiPromise, WsProvider, Keyring } from "@polkadot/api";
```
krzysztofziobro, line 1: Isn't that a similar thing to what e2e-tests `azero_to_eth` is doing?

relayer/relayer/src/contracts/azero.rs
```diff
@@ -35,9 +38,6 @@ pub enum AzeroContractError {
 
     #[error("Missing or invalid field")]
     MissingOrInvalidField(String),
-
-    #[error("Dry run failed")]
-    DryRunFailed(String),
```
krzysztofziobro, line 40: It actually would be best to dry-run here, so we avoid generating additional costs when crash-looping (also ensure that we dry run when submitting to Ethereum).

fbielejec: AFAIK dry running was not used anywhere but in `is_halted` for some reason, which is a query and therefore does not cost gas.  Should we want to start dry running this would have to be written from scratch for request signing, which is not a bad idea I guess. Question is if we want to do it in this PR or in a follow up?

azero/scripts/bridge_weth.ts
fbielejec: similar but not the same. This is a make target for quickly spawning requests when manually testing, I find it very usefull

relayer/relayer/src/listeners/azero.rs
fbielejec: indeed, when re-writing this part I forgot it waits for a whole era, this indeed is a problem. I'll think of a refactor that always proceeds but only moves the cache forward.

relayer/relayer/src/handlers/azero.rs
DamianStraszak: Yes, agreed with Krzysiek. The version with checking is infinitely more robust against virtually any problem.

relayer/relayer/src/handlers/azero.rs
```diff
@@ -0,0 +1,267 @@
...
+impl AlephZeroEventsHandler {
+    pub async fn run(
+        config: Arc<Config>,
+        eth_signed_connection: Arc<SignedEthConnection>,
+        mut azero_events_receiver: mpsc::Receiver<AzeroMostEvents>,
+        circuit_breaker_sender: broadcast::Sender<CircuitBreakerEvent>,
+        mut circuit_breaker_receiver: broadcast::Receiver<CircuitBreakerEvent>,
+    ) -> Result<CircuitBreakerEvent, AlephZeroEventsHandlerError> {
+        loop {
+            debug!("Ping");
```
DamianStraszak, line 183: If we want to keep these kind of logs, it's better to write more, because currently there are more places which log this exact same message.

DamianStraszak (changes requested):
Still reading it, so will add some comments later.

Looks infinitely better now. But still some small tweaks can be made.

relayer/relayer/src/handlers/azero.rs
fbielejec: they have the same message, but along that message they log the target (defaults to the module name), and that could be useful in checking if some process is e.g. deadlocked or refuses to die and blocks the restart, these types of problems

relayer/relayer/src/contracts/azero.rs
krzysztofziobro: We can do it in the follow-up if you prefer that

relayer/relayer/src/listeners/azero.rs
krzysztofziobro: But it processes batches in parallel, so most of the time, worst-case would be that transfer takes 2x more time, so it's acceptable + it would be a lot of work to rewrite it.
I think we can keep it that way

fbielejec: I think it's not that hard actually - we could:
- spawn independent tasks for batches of events
- use another channel to communicate when to seal a block
- channels are FIFO so the order will be preserved

relayer/relayer/src/contracts/azero.rs
fbielejec: yeah, I doubt I will have the time. It will be pointed out in the audit though, so we should probably mark it as out-of-scope as known. But if they see other gas savings this is within the scope

relayer/relayer/src/listeners/azero.rs
krzysztofziobro: If you see an easy way to do it, then I think it's worth to change it

azero/contracts/most/lib.rs
```diff
@@ -646,6 +652,30 @@ pub mod most {
             }
         }
 
+        #[ink(message)]
+        pub fn needs_signature(
+            &self,
+            request_hash: HashedRequest,
+            account: AccountId,
+        ) -> Result<bool, MostError> {
+            if !self.is_in_committee(self.get_current_committee_id()?, account) {
```
krzysztofziobro, line 661: This my fail for older requests which still need to be signed - remove that check or make needs_signature take correct committe_id as argument.

eth/contracts/Most.sol
```diff
@@ -317,10 +317,26 @@ contract Most is
...
-    ) external view returns (bool) {
+    ) public view returns (bool) {
         return pendingRequests[hash].signatures[guardian];
     }
 
+    function needsSignature(
+        bytes32 requestHash,
+        address account
+    ) external view returns (bool) {
+      if (!isInCommittee(committeeId, account)) {
```
krzysztofziobro, line 328: Same thing

relayer/relayer/src/handlers/azero.rs
krzysztofziobro: You added the check, but the main logic for checking if the change has finalized hasn't changed - please change it to the version in which check that changes happened is in form of call `needSigning` in the latest `Finalized` block. (wait_for_eth_tx_finality should not be needed)

krzysztofziobro (changes requested):

eth/contracts/Most.sol
fbielejec: changed

azero/contracts/most/lib.rs
fbielejec: changed

relayer/relayer/src/listeners/azero.rs
fbielejec: done

relayer/relayer/src/handlers/azero.rs
```diff
@@ -0,0 +1,281 @@
...
+                                Ok(_) => {},
+                                Err(why) => {
+                                    warn!("Event handler failed {why:?}, opening circuit breaker");
+                                    circuit_breaker_sender.send(CircuitBreakerEvent::AlephZeroEventHandlerFailure)?;
+                                },
+                            }
+                        }
+
+                        // marks the batch as done and notifies redis manager
+                        info!("Marking all events up to block {to_block} as handled");
```
krzysztofziobro, line 244: I don't think this is correct, as you spawn independent tasks per block, which means that you can complete processing of one block but something breaks in processing of the previous one a bit later - in that case redis value would be incorrect.

fbielejec: - we can assume that handling each batch take approximately the same amount of time, which is not an unreasonable assumption
- if we don't want to make that assumption we can make the channel have a buffer of 1

krzysztofziobro: We definitely don't want to make that assumption - empty blocks or ones that have all the transactions already processed can be processed much faster than other blocks (and there are scenarios where the previous blocks are more time-consuming to process).

I don't see how sending single event batches help here - their processing still can reorder.

What could work here is creating a one-shot for each batch of events and then receiving from these channels  sequentially in some select along other operation (in azero listener or redis manager, not sure). Not sure how easy to implement that is.

krzysztofziobro: I have no idea how such cache invalidation would be supposed to work - would you store the initial value in cache again and restart the listener? That's pretty inefficient, to make such mechanism you probably need to solve the original problem anyway

relayer/relayer/src/handlers/azero.rs
fbielejec: it's added

relayer/relayer/src/handlers/azero.rs
fbielejec: as discussed reverted to the version that awaits batch ack in the interest of time, we will devise something after the audit.

relayer/relayer/src/handlers/eth.rs
```diff
@@ -0,0 +1,206 @@
...
+            let amount = amount.as_u128();
+            let request_nonce = request_nonce.as_u128();
+
+            while contract
+                .needs_signature(
+                    azero_connection.as_connection(),
+                    request_hash,
+                    azero_connection.account_id().clone(),
+                    committee_id,
+                )
```
DamianStraszak, line 100: There is this question of which block is used to check this -- best or finalized? In theory if best is used (which it most likely is, but one would need to check), then this could actually behave incorrectly.

relayer/relayer/src/handlers/azero.rs
```diff
@@ -0,0 +1,249 @@
...
+
+                let address = eth_contract_address.parse::<Address>()?;
+                let contract = Most::new(address, eth_signed_connection.clone());
+
+                while contract
+                    .needs_signature(
+                        request_hash,
+                        eth_signed_connection.address(),
+                        committee_id.into(),
+                    )
```
DamianStraszak, line 112: Similarly as for the other half. It matters quite a bit which block is this checked against.

I would do the following:
loop {
    - check at most recent block, if `needs_signature`, then submit tx,
    - check at finalized block, if doesn't need signature, then break
}

DamianStraszak: If we always check only agains the most recent block, then this is not correct, as most recent block can revert

relayer/relayer/src/redis/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+        if *override_eth_cache {
+            write_block_number(
+                config.name.clone(),
+                ETH_BLOCK_KEY.to_string(),
+                redis_connection.clone(),
+                *config.default_sync_from_block_eth,
+            )?;
+        }
```
krzysztofziobro, line 73: This should not be done here, when it is and override_cache is specified, then it will reset the cache after each circuit break (or we need some "global" to guard for re-execution)

DamianStraszak (approved):

relayer/relayer/src/redis/mod.rs
fbielejec: [https://github.com/Cardinal-Cryptography/most/pull/131/commits/9f2c5d46e2412104191ec1d74019d79336138e84](https://github.com/Cardinal-Cryptography/most/pull/131/commits/9f2c5d46e2412104191ec1d74019d79336138e84)

relayer/relayer/src/redis/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+
+    #[error("channel receive error")]
+    Receive(#[from] broadcast::error::RecvError),
+}
+
+pub struct RedisManager;
+
+impl RedisManager {
+    pub async fn run(
+        is_first_run: Arc<AtomicBool>,
```
krzysztofziobro, line 40: No sure why you wouldn't pass just a plain `bool` here, it actually should be safer

krzysztofziobro (approved):

relayer/relayer/src/redis/mod.rs
fbielejec: if e.g. other components need to exclusively access it & alter - but I guess we can always change it

PR #132, opened 2024-03-07
Ink! contracts: fixing issues from internal audit

krzysztofziobro:
This PR aims to fix issues from the review:
- Adds a check for duplicates in the constructor and `set_committee`
- Removes callback in `set_code`
- Forbids calling `send_request` with `amount` 0
- Fixes issue in which a non-guardian could drain part of the rewards
- Changes fee calculation so that malicious oracle cannot cause an overflow
- Adds a requirement of MOST being halted for pair add/remove
- Removes `burn_from` from the token and makes `burn` permissioned.
- Check that ensures that MOST has minter permissions for the added token (in `add_pair`)

Additionally:
- Fixes issues with no pairs added in local development env
- Fixes e2e tests
- Changes constants used in tests and deployments on Aleph side to be reasonable

Changes to be addressed later:
- Type size changes for gas cost optimization
- Some simple tests (left to be implemented as drink tests to provide examples):
  * Duplicates in committee
  * Not allowing 0 amount in send_request
  * Not accepting token for which MOST has no minter permission


azero/contracts/token/lib.rs
```diff
@@ -227,11 +219,16 @@ pub mod token {
...
             Ok(())
         }
+
+        #[ink(message)]
+        fn minter(&self) -> AccountId {
+            self.minter
+        }
     }
 
     impl Burnable for Token {
```
krzysztofziobro, line 234: One thing that I'm not sure about: do we even want anyone but minter to be able to burn tokens?
Seems like it can cause a user error that causes the bridge to lose its invariant of having 1:1 correspondence between minted and locked tokens.

DamianStraszak: Yes, good point. I think maybe indeed only the "admin" should be able to burn.

DamianStraszak (approved):
Very nice.

azero/contracts/token/lib.rs
deuszx: I've always thought that the setup would be so that `addPair` on the MOST contract creates the instance of PSP22 token, setting itself as the "owner"/admin of that token.

DamianStraszak: Yes, it can be done this way too, but I think the current approach is OK -- the only serious downside is that it requires more manual steps.

DamianStraszak (approved):

PR #135, opened 2024-03-14
Examples of drink tests

krzysztofziobro:
- Detecting duplicates in updated committee
- Detecting 0 amount requests
- Test for outdated oracle price
- Test with correct `receive_request` execution


deuszx:
How about adding `drink-tests` to the CI?

deuszx (approved):
Code looks good, seems to compile, but we don't run it in CI - right? 

Nice job :)

PR #140, opened 2024-03-18
Add section about running tests to the README

krzysztofziobro:
Adding a description on how to run supported kinds of solidity and ink! tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,52 @@ Bring your assets to Aleph Zero and benefit from its scalable privacy-enhancing 
 
 Have docker and docker-compose installed.
 
+## Testing the contracts
+
+### Solidity contracts
+
+To run the tests for the Solidity `Most.sol` contract, run:
+
+```bash
+make test-solidity
+```
+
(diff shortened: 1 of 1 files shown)


obrok (changes requested):
These all work for me except for `ink_e2e tests` - perhaps I'm doing something wrong, but just doing `make test-ink-e2e` gives errors like:

```
test e2e::base_fee_too_low ... thread 'e2e::base_fee_too_low' panicked at /home/yapee/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ink_e2e-4.3.0/src/node_proc.rs:192:10:
We should find a port before the reader ends
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
FAILED
```

I tried doing `make bridge` first, no joy. Am I missing some prerequisite? If so, it should be describe in this README too.

obrok (approved):

krzysztofziobro:
You need to have `docker-compose` in version `1.29.2` installed

PR #141, opened 2024-03-19
A0-4114: Add deployment to testnet

mikogs:


obrok (changes requested):
I don't think it makes any sense to run `deploy-eth`/`deploy-azero` for testnet at this point, unless you aim to create a "private" instance of MOST - not the one for the internal/guardian deployment with this.

PR #142, opened 2024-03-20
Adjust deployment scripts for testnet

obrok:
* Adds a completely new script for ETH - the existing one is entangled with Gnosis Safe deployment, which I don't think we need.
* Tweaks the Aleph Zero deployment script so that it can work with existing tokens with known addresses.

Instructions for deployment with these:

* Prepare config based on azero/env/testnet.json.example -> azero/env/testnet.json
* Adjust deploymentConfig for sepolia in eth/hardhat.config.js
* AZERO_ENV=testnet SEPOLIA_KEY=[private key] NETWORK=sepolia make deploy-live
* Get the address and verify the contract on etherscan
    * SEPOLIA_KEY=[private key] NETWORK=sepolia CONTRACT=[address] make verify-eth

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,6 +105,10 @@ deploy-eth-live: compile-eth
 	cd eth && \
 	npx hardhat run --network $(NETWORK) scripts/deploy_bridge_live.js
 
+.PHONY: deploy-live
+deploy-live: # Deploy azero and eth contracts on a live network (testnet or mainnet)
+deploy-live: deploy-azero setup-azero deploy-eth-live
+
 .PHONY: verify-eth
 verify-eth: # Post verified eth sources of a contract to etherscan
 verify-eth:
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -99,6 +99,18 @@ deploy-eth: compile-eth
 	npx hardhat run --network $(NETWORK) scripts/1_deploy_gnosis_safe.js && \
 	npx hardhat run --network $(NETWORK) scripts/2_deploy_bridge_contracts.js
 
+.PHONY: deploy-eth-live
```
krzysztofziobro, line 102: Would be nice to add one make instruction that just takes mnemonics (or they could be input to configs before deployment, both are OK) and deploys all contracts on testnets - now it's not immediately clear how to do it.

krzysztofziobro (approved):
nice

PR #147, opened 2024-03-24
Relayer improvements

krzysztofziobro:
- Fixes e2e tests and tx submission scripts
- Adds dry-running before submitting to the chain
- Improves tx submission loop in azero handler 
- Parallel handling of azero block batches

### DIFF
--- a/relayer/relayer/src/contracts/azero.rs
+++ b/relayer/relayer/src/contracts/azero.rs
@@ -173,7 +173,7 @@ impl MostInstance {
             .await??)
     }
 
-    pub async fn needs_signature(
+    pub async fn _needs_signature(
         &self,
         connection: &Connection,
         request_hash: [u8; 32],
--- a/relayer/relayer/src/contracts/azero.rs
+++ b/relayer/relayer/src/contracts/azero.rs
@@ -13,11 +13,11 @@ use aleph_client::{
         ContractMessageTranscoder,
         Value::{self, Seq},
     },
-    pallets::contract::ContractsUserApi,
+    pallets::contract::{ContractCallArgs, ContractRpc, ContractsUserApi},
     sp_weights::weight_v2::Weight,
-    AccountId, AlephConfig, Connection, TxInfo, TxStatus,
+    AccountId, AlephConfig, Connection, SignedConnectionApi, TxInfo, TxStatus,
 };
-use log::trace;
+use log::{error, trace};
 use subxt::events::Events;
 use thiserror::Error;
 
@@ -38,6 +38,12 @@ pub enum AzeroContractError {
 
     #[error("Missing or invalid field")]
     MissingOrInvalidField(String),
+
+    #[error("Dry-run reverted")]
+    DryRunReverted(Result<Value, anyhow::Error>),
+
+    #[error("Dispatch error")]
+    DispatchError(String),
 }
 
 pub struct AdvisoryInstance {
@@ -107,6 +113,10 @@ impl MostInstance {
         dest_receiver_address: [u8; 32],
         request_nonce: u128,
     ) -> Result<TxInfo, AzeroContractError> {
+        let gas_limit = Weight {
+            ref_time: self.ref_time_limit,
+            proof_size: self.proof_size_limit,
+        };
         let args = [
             bytes32_to_str(&request_hash),
             committee_id.to_string(),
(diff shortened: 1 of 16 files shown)


relayer/relayer/src/contracts/azero.rs
```diff
@@ -38,6 +38,12 @@ pub enum AzeroContractError {
 
     #[error("Missing or invalid field")]
     MissingOrInvalidField(String),
+
+    #[error("Dry-run reverted")]
+    DryRunReverted(Result<Value, anyhow::Error>),
```
fbielejec, line 43: it's a bit weird to return Result in an Err variant. Maybe ditch the `anyhow` part as it contains no useful information anyway

relayer/relayer/src/contracts/azero.rs
```diff
@@ -140,7 +173,7 @@ impl MostInstance {
             .await??)
     }
 
-    pub async fn needs_signature(
+    pub async fn _needs_signature(
```
fbielejec, line 176: checking this could still save guardian some gas if request collected enough signatures before he committed his

relayer/relayer/src/contracts/azero.rs
DamianStraszak: Impossible to change it, the way I see it, because it's in a crate that we don't control, and certainly don't want to fork it just for that currently.

DamianStraszak: Unless you mean to return just `anyhow::Error`? This is probably fine.

DamianStraszak (approved):

relayer/relayer/src/contracts/azero.rs
krzysztofziobro: We currently have no way of calling this for a finalized block, only for latest so using this is unsafe unfortunately.

relayer/relayer/src/handlers/eth.rs
```diff
@@ -91,39 +89,26 @@ impl EthereumEventHandler {
             let amount = amount.as_u128();
             let request_nonce = request_nonce.as_u128();
 
-            while contract
-                .needs_signature(
-                    azero_connection.as_connection(),
+            contract
```
fbielejec, line 92: looping is indeed probably not needed here, but the guardian could still confirm if his signature is needed before signing the request

krzysztofziobro: He cannot because we cannot make this call for the last finalized block, so we could omit a tx that will actually need our signature.

relayer/relayer/src/listeners/azero.rs
```diff
@@ -65,8 +69,8 @@ impl AlephZeroListener {
         config: Arc<Config>,
         azero_connection: Arc<Connection>,
         azero_events_sender: mpsc::Sender<AzeroMostEvents>,
-        last_processed_block_number: broadcast::Sender<u32>,
-        mut next_unprocessed_block_number: broadcast::Receiver<u32>,
+        next_block_to_process_send: broadcast::Sender<u32>,
```
fbielejec, line 72: `next_block_to_process_sender` to keep variable naming consistent

relayer/relayer/src/listeners/azero.rs
```diff
@@ -65,8 +69,8 @@ impl AlephZeroListener {
...
+        mut next_block_to_process_recv: broadcast::Receiver<u32>,
```
fbielejec, line 73: `next_block_to_process_receiver`

relayer/relayer/src/handlers/eth.rs
```diff
@@ -91,39 +89,26 @@ impl EthereumEventHandler {
...
-                        dest_token_address: hex::encode(dest_token_address),
-                        amount,
-                        dest_receiver_address: hex::encode(dest_receiver_address),
-                        request_nonce,
-                    })?;
-
-                sleep(Duration::from_secs(ALEPH_BLOCK_PROD_TIME_SEC)).await;
-            }
+                .await
+                // default AlephClient error is MBs large and useless, dumps the entire runtime for some reasons
```
fbielejec, line 103:
```suggestion
                // default AlephClient error is MBs large and useless, dumps the entire runtime for some reason
```

relayer/relayer/src/redis/mod.rs
```diff
@@ -110,7 +110,7 @@ impl RedisManager {
                         name.clone(),
                         ETH_BLOCK_KEY.to_string(),
                         Arc::clone(&redis_connection),
-                        seal_block_number
+                        seal_block_number + 1
```
fbielejec, line 113: question is what do we want this value to be - last processed block or the next block to process. Next block makes sense because that is how the relayer works rn, but conceptually storing last processed block makes sense as well - for manual inspection, or if some other software needs to read it.  Small price to pay is if relayer crashed he goes through that block again, but it's harmless

krzysztofziobro: From how the code uses it - it should be next block to process and I think it's ok to keep it that way

relayer/relayer/src/handlers/azero.rs
```diff
@@ -239,6 +261,9 @@ impl AlephZeroEventsHandler {
                             }
                         }
 
+                        if ack.send(to_block).is_err() {
```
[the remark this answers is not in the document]
krzysztofziobro: It would need to be for "error" of oneshot sender, which is not a meaningful error

PR #153, opened 2024-03-28
Docs

fbielejec:


docs/relayer_hldd.org
```diff
@@ -0,0 +1,112 @@
...
+- *Handlers*: independent event handlers that are spawned for each request. After all the events from the batch of blocks acknowledge they have finished cache gets updated with the next block number.
+Handlers sign and submit transactions to the MOST smart contract on the Ethereum chain, first by checking whether a given guardian signature is needed (maybe she already signed it, or maybe the request is already processed, or maybe some previous blocks were yanked). As long as it is needed in the last finalized block on ethereum, handler will sign it and wait for as long as the transaction appears in the last finalized block. Following pseudo-code is a conceptual representation of the exact logic followed by the handlers before acknowleding an cross-chain transfer request as handled:
```
obrok, line 59:
```suggestion
Handlers sign and submit transactions to the MOST smart contract on the Ethereum chain, first by checking whether a given guardian signature is needed (maybe they already signed it, or maybe the request is already processed, or maybe some previous blocks were yanked). As long as it is needed in the last finalized block on ethereum, the handler will sign it and wait until the transaction appears in the last finalized block. The following pseudo-code is a conceptual representation of the exact logic followed by the handlers before acknowledging a cross-chain transfer request as handled:
```

docs/relayer_hldd.org
```diff
@@ -0,0 +1,112 @@
...
+eth_listener .up.> eth_events : publish batch
+eth_listener <..> eth_block_number: pub-sub
+eth_handler <.up. eth_events : subsciption
+eth_handler ..> a0 : tx
+redis <.. eth_block_number: subscription
+redis ..> eth_block_number: publish on boot
+#+END_SRC
+[[file:eth_azero_components.png][eth_azero_components.png]]
+
+- *Listener*: maintains a subscription to the Ethereum chain. On boot it receives next block number after the last cached and proceeds to retrieve events in batches coming from consecutive blocks, between the last processed and the last finalized, whichever number is smaller. For each batch *Listener* publishes it on an Events channel for a *Handler* to pick up and process in a blokciing manner, i.e. the *Listener* doe snot process further until the current batch is entirely handed and the ecache is updated.
```
obrok, line 103:
```suggestion
- *Listener*: maintains a subscription to the Ethereum chain. On boot it receives the next block number after the last one cached and proceeds to retrieve events in batches coming from consecutive blocks, between the last processed and the last finalized, whichever number is smaller. For each batch, *Listener* publishes it on an Events channel for a *Handler* to pick up and process in a blokcing manner, i.e. the *Listener* does not process further until the current batch is entirely handled and the cache is updated.
```

docs/relayer_hldd.org
```diff
@@ -0,0 +1,112 @@
...
+- *Handler*: a task which subscribes to the Events channels and acks whenever a given batch of transfer requests is succesfully handled. Each event in a batch is processed sequentially in a FIFO order.
```
obrok, line 104:
```suggestion
- *Handler*: a task which subscribes to the Events channels and acks whenever a given batch of transfer requests is successfully handled. Each event in a batch is processed sequentially in a FIFO order.
```

docs/relayer_hldd.org
```diff
@@ -0,0 +1,112 @@
...
+
+* DONE Additional components
+There are various ancillary components not part of the diagrams above:
+
+- *Circuit Breaker*: Ciruit Breaker is a multiple-producer multiple-consumer channel, a message pushed to this channels by one of the components makes the remaining components drop all their tasks immediately, which returns control to the main thread. The relayer enters a shutdown/reboot loop, in which state the system tries to handle the next batch of requests, starting from the last succesfully processed block until the condition is removed.
```
obrok, line 109:
```suggestion
- *Circuit Breaker*: Ciruit Breaker is a multiple-producer multiple-consumer channel, a message pushed to this channels by one of the components makes the remaining components drop all their tasks immediately, which returns control to the main thread. The relayer enters a shutdown/reboot loop, in which state the system tries to handle the next batch of requests, starting from the last successfully processed block until the condition is removed.
```

docs/relayer_hldd.org
```diff
@@ -0,0 +1,112 @@
...
+- *Advisory Listener*: this is a process that listens to none, one or a collection of Advisory smart contracts, which are boolean flag contracts, deployed on the AlephZero chain, where interested parties can publish warnings that shut down the relayer operation. Notice that the smart contracts comprising the MOST do not observe these flags, only the relayer process does. If the Advisory Listener detects a flag it publishes to the circuit breaker channel.
```
obrok, line 110:
```suggestion
- *Advisory Listener*: this is a process that listens to none, one or a collection of Advisory smart contracts, which are boolean flag contracts, deployed on the AlephZero chain, where interested parties can publish warnings that shut down the relayer operation. Notice that the smart contracts comprising the MOST do not observe these flags, only the relayer process does. If the Advisory Listener detects a flag it publishes to the circuit breaker channel.
```

docs/relayer_hldd.org
```diff
@@ -0,0 +1,112 @@
...
+- *Halted state Listener*: Twin processes that check whether the bridge on ther side has beeen halted by a governance ation. Detecting it means an event is published to the circuit breaker channel.
```
obrok, line 111:
```suggestion
- *Halted state Listener*: Twin processes that check whether the bridge on their side has beeen halted by a governance action. Detecting it means an event is published to the circuit breaker channel.
```

docs/relayer_hldd.org
```diff
@@ -0,0 +1,112 @@
...
+- *Signer*: a separate process reponsible for signing cross-chain request transaction payloads with a key stored within a hardware enclave.
```
obrok, line 112:
```suggestion
- *Signer*: a separate process responsible for signing cross-chain request transaction payloads with a key stored within a hardware enclave.
```

docs/relayer_hldd.org
fbielejec: there is a mistake in the description, see latest changes

obrok: Have you pushed them yet? I don't seem to see anything new.

docs/relayer_hldd.org
fbielejec: done

docs/relayer_hldd.org
obrok: :+1:

fbielejec:
> I like that the source for the diagrams is embedded in the same file. I'd suggest mentioning this doc in the README and explaining how to view it (I guess just open on github, should render it?) or compile it on your machine. If you do so, then I think you can remove the compiled diagrams from the repo.

GH renders .org with the png, but I don't think it can compile plantuml diagrams - I included instructions on rendering them.

PR #161, opened 2024-04-08
Fix/client connection error

fbielejec:


relayer/relayer/src/main.rs
```diff
@@ -217,15 +224,16 @@ async fn main() -> Result<(), RelayerError> {
...
 
-fn run_relayer(
+async fn run_relayer(
     first_run: bool,
     tasks: &mut JoinSet<Result<CircuitBreakerEvent, RelayerError>>,
     config: Arc<Config>,
-    azero_connection: Arc<AzeroWsConnection>,
-    azero_signed_connection: Arc<AzeroConnectionWithSigner>,
     eth_connection: Arc<EthConnection>,
     eth_signed_connection: Arc<SignedEthConnection>,
```
krzysztofziobro, line 232: I think it would be best to recreate all the connections each time - we haven't encountered it, but some issue might arise with eth connections which also requires recreating them

fbielejec: Done

PR #163, opened 2024-04-10
Feature/payout accounts

fbielejec:


azero/contracts/most/lib.rs
```diff
@@ -708,6 +722,29 @@ pub mod most {
             Ok(())
         }
 
+        /// Sets address of the account that receives rewards on behalf of the committee member.
+        ///
+        /// Can only be called by an account that was a committee member in `committee_id`.
+        /// All the unpaid rewards will then be paid out to this account address when `payout_request` is called.
+        /// Comittee members that have set the `payout_account` do not need to call this method again when the committee changes, unless they want to change the payouts account.
+        #[ink(message)]
+        pub fn set_payout_account(
```
krzysztofziobro, line 731: I think a simpler approach would be to just add an argument to `payout_rewards`, which would be a target account of the payout.

With the current approach we add additional message + handling fees in relayer will need to call two separate methods instead of just one.

fbielejec: we can't - AFAIK payout_rewards is callable by anyone

azero/contracts/most/lib.rs
```diff
@@ -464,22 +477,24 @@ pub mod most {
...
-        /// Can be called by anyone on behalf of the committee member.
+        /// Reverts if `member_id` account is not in the committee with `committee_id`
         #[ink(message)]
         pub fn payout_rewards(
             &mut self,
             committee_id: CommitteeId,
             member_id: AccountId,
         ) -> Result<(), MostError> {
             self.ensure_not_halted()?;
+            self.only_committee_member(committee_id, member_id)?;
```
krzysztofziobro, line 488: Now this check seems unnecessary

fbielejec: an early return, saves gas - but yeah we can remove. Let me know with :+1: / :-1:

krzysztofziobro: Ah, ok, makes sense, we ca leave it as is

azero/contracts/most/lib.rs
```diff
@@ -744,6 +760,29 @@ pub mod most {
             Ok(())
         }
 
+        /// Sets address of the account that receives rewards on behalf of the committee member.
+        ///
+        /// Can only be called by an account that was a committee member in `committee_id`.
+        /// All the unpaid rewards will then be paid out to this account address when `payout_request` is called.
+        /// Comittee members that have set the `payout_account` do not need to call this method again when the committee changes, unless they want to change the payouts account.
+        #[ink(message)]
```
krzysztofziobro, line 768: I think it would be good to also have a getter for this

krzysztofziobro (approved):

PR #166, opened 2024-04-12
Do not crash relayer on old committee requests

krzysztofziobro:
Makes relayer omit the requests which do not belong to his committee instead of crashing - useful when rotating new guardians into the committee.


relayer/relayer/src/handlers/azero.rs
```diff
@@ -183,6 +194,22 @@ impl AlephZeroEventHandler {
...
+    most: &Most<SignedEthConnection>,
+    committee_id: U256,
+    address: Address,
+) -> Result<bool, AlephZeroEventHandlerError> {
+    if most.is_in_committee(committee_id, address).await? {
+        return Ok(false);
+    }
+
+    if committee_id > most.committee_id().await? {
+        error!("Request from a future committee {committee_id} - this likely indicates MOST contracts misconfiguration");
```
fbielejec, line 207: if you;re not crashing log level should be  <= `warn`

relayer/relayer/src/handlers/eth.rs
```diff
@@ -117,6 +123,33 @@ impl EthereumEventHandler {
...
+            connection.as_connection(),
+            committee_id,
+            connection.account_id().clone(),
+        )
+        .await?
+    {
+        return Ok(false);
+    }
+
+    if committee_id
```
fbielejec, line 142: Can this even happen? Relayer should not keep on running if it does

fbielejec (changes requested):

relayer/relayer/src/handlers/eth.rs
krzysztofziobro: It will crash when executing the request - it shouldn't happen if the rotation is performed on both ends of the bridge, only happens if governance makes a mistake

relayer/relayer/src/handlers/azero.rs
DamianStraszak: To me this also seems like a case where you can return an error and crash/(trigger circuit breaker). It is certainly unexpected that you see a request from a committee that you don't know. This would mean there is a problem with the bridge, or you are not synced.

relayer/relayer/src/handlers/azero.rs
```diff
@@ -107,6 +107,17 @@ impl AlephZeroEventHandler {
...
                 let contract = Most::new(address, eth_signed_connection.clone());
 
+                if past_request(
+                    &contract,
+                    committee_id.into(),
+                    eth_signed_connection.address(),
+                )
+                .await?
+                {
+                    info!("Guardian signature for {request_hash:?} not needed - request from a past committee");
```
DamianStraszak, line 117: Strictly formally -- it's not necessarily from "past committee", it's from a committee we are not part of. 

In fact I would consider changing the name `past_request` to `not_in_committee`.

krzysztofziobro: True, will rename

relayer/relayer/src/handlers/eth.rs
fbielejec: I see - yet still best to fail early if you can

DamianStraszak (approved):

relayer/relayer/src/handlers/eth.rs
```diff
@@ -89,6 +93,11 @@ impl EthereumEventHandler {
             let amount = amount.as_u128();
             let request_nonce = request_nonce.as_u128();
 
+            if not_in_committee(&contract, azero_connection, committee_id).await? {
+                info!("Guardian signature for {request_hash:?} not needed - request from a past committee");
```
fbielejec, line 97:
```suggestion
                info!("Guardian signature for {request_hash:?} not needed - request from a different committee");
```

fbielejec (commented):
log message suggestion, otherwise LGTM and good for merging

fbielejec (approved):

PR #168, opened 2024-04-16
Remove Unexpected error that crashes the relayer when bridge is paused

krzysztofziobro:


relayer/relayer/src/listeners/eth.rs
```diff
@@ -221,25 +218,40 @@ impl EthereumPausedListener {
...
+
+            select! {
+                cb_event = circuit_breaker_receiver.recv () => {
+                    warn!(target: "EthereumPausedListener","Exiting due to a circuit breaker event {cb_event:?}");
+                    return Ok(cb_event?);
+                },
 
-            _ = async {
-                loop {
+                is_paused = is_paused_call.call() => {
```
fbielejec, line 232: isn't a call for transacting, doesn't this cost gas?

krzysztofziobro: `send` is for transacting, `call` for read-only/dry-run

fbielejec (approved):

PR #170, opened 2024-04-18
Supported pairs query

fbielejec:
Adding a missing query on the azero contract ( supported pairs )


JanKuczma:
It's not a part of this PR but maybe it could be fixed here as it touches [bridge_weth.ts](https://github.com/Cardinal-Cryptography/most/blob/master/azero/scripts/bridge_weth.ts) file:
https://github.com/Cardinal-Cryptography/most/blob/9a35d3ae20f4ed0411a22cd1f051fb0d14bdc768/azero/scripts/bridge_weth.ts#L14-L20
```ts
function getTokenAddresses(addresses, tokenName): [string, string] {
  for (let [_tokenName, ethAddress, azeroAddress] of addresses.tokens) {
    if (tokenName == _tokenName) {
      return [ethAddress, azeroAddress];
    }
  }
}
```
https://github.com/Cardinal-Cryptography/most/blob/9a35d3ae20f4ed0411a22cd1f051fb0d14bdc768/azero/scripts/bridge_weth.ts#L28
```ts
  const weth_azero = getTokenAddresses(addresses, "wETH")[1];
```

fbielejec:
Can we merge this @krzysztofziobro @obrok ? Best before mainnet release I suppose?

krzysztofziobro (commented):
I would merge it, but I think it would be nice to add here also getters for fee constants + I would remove the upgrade script (upgrades need to be manual performed by the contract owner anyway).

fbielejec:
> I would merge it, but I think it would be nice to add here also getters for fee constants + I would remove the upgrade script (upgrades need to be manual performed by the contract owner anyway).

Done, please double check me if we're not missing sth

krzysztofziobro (approved):

PR #172, opened 2024-04-22
Fix bridging ether

JanKuczma:
- Change the implicit bridging weth(azero) -> ether(eth) to be explicit through `send_request_native_ether(...)`
- Allow bridging weth(azero) to weth(eth) through `send_request(...)`

### DIFF
--- a/azero/contracts/most/lib.rs
+++ b/azero/contracts/most/lib.rs
@@ -209,6 +209,7 @@ pub mod most {
         HaltRequired,
         NoMintPermission,
         ZeroAddress,
+        WrappedEthNotSet,
     }
 
     impl From<InkEnvError> for MostError {
@@ -403,7 +404,7 @@ pub mod most {
         ) -> Result<(), MostError> {
             self.ensure_not_halted()?;
 
-            let src_token_address = self.weth.get().ok_or(MostError::UnsupportedPair)?;
+            let src_token_address = self.weth.get().ok_or(MostError::WrappedEthNotSet)?;
 
             // ETH_ZERO_ADDRESS as `dest_token_address` indicates native ether transfer
             self._send_request(
--- a/azero/contracts/most/lib.rs
+++ b/azero/contracts/most/lib.rs
@@ -156,8 +156,6 @@ pub mod most {
         gas_price_oracle: Option<AccountId>,
         /// Is the bridge in a halted state
         is_halted: bool,
-        /// Wrapped ethereum (azero) address. Necessary to perform bridging weth(azero) -> ether(eth).
-        weth: Option<AccountId>,
     }
 
     #[ink(storage)]
@@ -185,6 +183,8 @@ pub mod most {
         paid_out_member_rewards: Mapping<(AccountId, CommitteeId), u128, ManualKey<0x50414944>>,
         /// committe members can specify a special account for collecting the rewards, different from the one used for signing
         payout_accounts: Mapping<AccountId, AccountId>,
+        /// Wrapped ethereum (azero) address. Necessary to perform bridging weth(azero) -> ether(eth).
+        weth: Lazy<AccountId, ManualKey<0x7CD95FED>>,
     }
 
     #[derive(Debug, PartialEq, Eq, Encode, Decode)]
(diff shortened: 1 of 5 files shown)


azero/contracts/most/lib.rs
```diff
@@ -156,6 +156,8 @@ pub mod most {
         gas_price_oracle: Option<AccountId>,
         /// Is the bridge in a halted state
         is_halted: bool,
+        /// Wrapped ethereum (azero) address. Necessary to perform bridging weth(azero) -> ether(eth).
+        weth: Option<AccountId>,
```
krzysztofziobro, line 160: Question: shouldn't this be in a separate `Lazy` to ensure a smooth upgrade on testnet?

azero/contracts/most/lib.rs
```diff
@@ -374,6 +377,79 @@ pub mod most {
...
+        pub fn send_request_native_ether(
+            &mut self,
+            amount: u128,
+            dest_receiver_address: [u8; 32],
+        ) -> Result<(), MostError> {
+            self.ensure_not_halted()?;
+
+            let mut data = self.data()?;
+
+            let src_token_address = data.weth.ok_or(MostError::UnsupportedPair)?;
```
krzysztofziobro, line 394: I think we may want to add a custom error here, not just `UnsupportedPair`, which is inaccurate

azero/contracts/most/lib.rs
```diff
@@ -374,6 +377,79 @@ pub mod most {
...
+            data.request_nonce = request_nonce.checked_add(1).ok_or(MostError::Arithmetic)?;
+
+            self.data.set(&data);
+
+            // return surplus if any
+            if let Some(surplus) = transferred_fee.checked_sub(current_base_fee) {
+                if surplus > 0 {
+                    self.env().transfer(sender, surplus)?;
+                }
+            };
```
krzysztofziobro, line 440: Could you try to extract some parts of that code so it can be reused in both send_requests?

krzysztofziobro (commented):
Would be cool if you could also add some basic test for the `send_request_native_ether` (for example in drink!)

azero/contracts/most/lib.rs
JanKuczma: I think `UnsupportedPair` is not entirely inaccurate in the context of `send_request_native_ether(...)` as it suggests that the "pair" ether(azero) -> ether(eth) has not been set (just like in the context of `send_requst(...)` for regular pairs).
However, to be more explicit we can replace it with a custom error like `WrappedEthNotSet`.

azero/contracts/most/lib.rs
JanKuczma: Definitely

azero/contracts/most/lib.rs
krzysztofziobro: Yes, I agree that the issue is somewhat similar to the one typically indicated by UnsuppoetedPair, but I think it's different enough to get it's own error variant

azero/contracts/most/lib.rs
JanKuczma: You are correct. An upgrade would break the `data: Lazy<Data,...>` because `weth` is not initialized. It should be moved to a separate `Lazy`.

fbielejec: you could add a migration that copies all the old data and sets the new field:
```
            if let Some(OldData { x, y, .. }) = get_contract_storage(&123)? {               
                self.data.set(&Data { x: x, y: y, z: weth_address });
                return Ok(());
            }
```

fbielejec (commented):
I would suggest:

- either changing `weth` to be an `Option`
- .. or merging it with the Data (like in the first version) and adding a migration that sets it

azero/contracts/most/lib.rs
```diff
@@ -278,6 +281,7 @@ pub mod most {
 
             let mut ownable_data = Lazy::new();
             ownable_data.set(&Ownable2StepData::new(owner));
+            let weth = Lazy::new();
```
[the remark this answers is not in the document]
krzysztofziobro: Making it Lazy and putting under a manual key should make future upgrades easier (or at least more consistent)

azero/contracts/most/lib.rs
JanKuczma: This way we need to keep the definition of the "`OldData`" just for the process of migration. I agree there is no need for the `set_weth(...)` but instead we need `migrate(...)`. This introduces more complexity imho.

krzysztofziobro (approved):

PR #175, opened 2024-04-25
Nonce handling by hand

kostekIV:


.github/workflows/on-master-branch-commit.yml
```diff
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - master
+      - nonce
```
fbielejec, line 8: don;t forget to remove!

relayer/relayer/Cargo.toml
```diff
@@ -23,3 +23,4 @@ signer_client = { path = "../signer_client" }
 subxt = { workspace = true }
 thiserror = "1.0.48"
 tokio = { workspace = true, features = ["rt", "rt-multi-thread", "macros", "test-util", "sync"] }
+tracing = { workspace = true, features = ["log", "log-always"] }
```
fbielejec, line 26: I'm adding it in another PR,  but cool, lets keep it here

.github/workflows/on-master-branch-commit.yml
kostekIV: done

relayer/relayer/src/connections/azero.rs
```diff
@@ -1,3 +1,5 @@
+use std::sync::{Arc, Mutex as StdMutex};
```
fbielejec, line 1: I would called SyncMutex (since Tokio is Async/await)

kostekIV: done

relayer/relayer/src/connections/azero.rs
```diff
@@ -1,3 +1,5 @@
+use std::sync::{Arc, Mutex as SyncMutex};
```
krzysztofziobro, line 1: Is there some good reason here to use Mutex from std instead of tokio Mutex?

kostekIV: 2nd case I think we should only increase the nonce if we were successful in submitin it to the chain. I can use the tokio mutex, np

fbielejec (approved):
You could remove some of the duplication by creating a `with_nonce` decorator that augments the existing type, instead of adding the nonce to the `with_keypair`, but generally LGTM

relayer/relayer/src/connections/azero.rs
fbielejec: I think the sync version has less overhead, even in async scenarios (read some articles on that).

relayer/relayer/src/connections/azero.rs
kostekIV: changed to atomicu64

relayer/relayer/src/connections/azero.rs
kostekIV: after discusion changed it to Atomicu64

kostekIV:
> You could remove some of the duplication by creating a `with_nonce` decorator that augments the existing type, instead of adding the nonce to the `with_keypair`, but generally LGTM

I dont like macros etc :P So I have just created new common method.

fbielejec:
> > You could remove some of the duplication by creating a `with_nonce` decorator that augments the existing type, instead of adding the nonce to the `with_keypair`, but generally LGTM
> 
> I dont like macros etc :P So I have just created new common method.

You could do it w-out any macros.. but this works equally fine

PR #177, opened 2024-04-26
A0-4251: Avoid resumbissions on Aleph via better handling loop

ggawryal:
Improved loop in ETH handler based on Azero handler. For this, I have introduced some changes in aleph-client: https://github.com/Cardinal-Cryptography/aleph-node/pull/1679.


e2e-tests/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-aleph_client = { git = "https://github.com/Cardinal-Cryptography/aleph-node/", rev = "d52ede9" }
+aleph_client = { git = "https://github.com/Cardinal-Cryptography/aleph-node/", rev = "c3c6d8b37d586b6b050d1e40675dd6b4ab6adcdb" }
```
kostekIV, line 7: just a heads up. You probably want to cherry-pick changes to aleph-client into release-13. Othwewise relayer build upon this pr wont be working before the next testnet release.

relayer/Cargo.toml
```diff
@@ -3,7 +3,7 @@ members = ["relayer", "signer_client", "signer", "signer_tester"]
 resolver = "2"
 
 [workspace.dependencies]
-aleph_client = { git = "https://github.com/Cardinal-Cryptography/aleph-node", rev = "r-13.2" }
+aleph_client = { git = "https://github.com/Cardinal-Cryptography/aleph-node", rev = "c3c6d8b37d586b6b050d1e40675dd6b4ab6adcdb" }
```
kostekIV, line 6: and I meant the comment to be here actually

relayer/relayer/src/contracts/azero.rs
```diff
@@ -125,44 +120,12 @@ impl MostInstance {
...
-
-        let call_result = signed_connection
-            .call(
-                self.address.clone(),
-                0,
-                gas_limit,
-                None,
-                call_data,
-                TxStatus::Finalized,
-            )
```
ggawryal, line 165: I have quite a strong opinion for migrating to subxt eventually, so I can leave explicit dry-running and `call` instead of the higher level `exec` from aleph-client, to facilitate this in the future. Wdyt?

e2e-tests/Cargo.toml
ggawryal: Will update once https://github.com/Cardinal-Cryptography/aleph-node/pull/1704 is merged

relayer/relayer/src/contracts/azero.rs
```diff
@@ -172,27 +135,36 @@ impl MostInstance {
...
     }
 
-    pub async fn _needs_signature(
+    pub async fn needs_signature(
         &self,
         connection: &Connection,
         request_hash: [u8; 32],
         account: AccountId,
         committee_id: u128,
+        block: BlockStatus,
```
fbielejec, line 148: I don't see a situation where we would need the make a query at a block other than the finalized.

fbielejec (approved):
looks good, nice work

PR #178, opened 2024-05-02
Trader component

fbielejec:
- opt-in component
- should never soft nor hard crash the relayer during operation

To invoke it agains testnet use these addresses:
```
ARGS=(
  run --bin relayer -- --name "guardian_${RELAYER_ID}" \
  --advisory-contract-addresses="5HWVsS6bzxdW7qWYVuwvaRyHKTTkPq7hNEjaRrDpv9QTupm1" \
  --azero-contract-address="5Csn41URV2L66quA2pzAQKQKv2EAYbUL5z58MT1rbVRGpkVW" \
  --eth-contract-address="0x066A9ac7678570fa88d672724DDc545292373401" \
  --eth-node-http-url="https://sepolia.infura.io/v3/<key>" \
  --azero-node-wss-url="wss://ws.test.azero.dev:443" \
  --dev-account-index=${RELAYER_ID} \
  --default-sync-from-block-eth=5531318 \
  --default-sync-from-block-azero=57570232 \
  --run-trader-component \
  --router-address=5FmVtQGGUDVSpqYVkU3zLBQProQ4hHPjGQjUahxCYLJJvjfK \
  --azero-wrapped-azero-address=5EFDb7mKbougLtr5dnwd5KDfZ3wK55JPGPLiryKq4uRMPR46 \
  --azero-ether-address=5GCESNyxspadqV267XA1jr1dUs5gQpM5fK8vWf2mbUCHrNcX \
  --eth-wrapped-ether-address=0xd91aE8FD2Be53F74876a9cc4aFb416645A0c8420 \
  --dev
)
```


relayer/relayer/src/contracts/azero.rs
```diff
@@ -97,6 +226,35 @@ impl MostInstance {
...
+            ref_time: self.ref_time_limit,
+            proof_size: self.proof_size_limit,
+        };
+        let args = [
+            bytes32_to_str(&src_token_address),
+            amount.to_string(),
+            bytes32_to_str(&dest_receiver_address),
+        ];
+
+        let params = ExecCallParams::new().gas_limit(gas_limit);
```
fbielejec, line 246: TODO: Add base_fee

relayer/relayer/src/trader/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+                    let azero_balance = azero_signed_connection
+                        .get_free_balance(whoami_azero.to_owned(), None)
+                        .await;
+
+                    info!("{whoami_azero} has a balance of: {azero_balance} Azero.");
+
+                    let current_base_fee = match most_azero.get_base_fee(azero_signed_connection.as_connection()).await {
+                        Ok(amount) => {
+                            info!("Current base fee: {amount} pA0");
+                            amount},
```
deuszx, line 150: Weird formatting - did you forget to run `cargo fmt`?

relayer/relayer/src/trader/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+                            warn!("Query to `get_base_fee` has failed {why:?}.");
+                            continue;
+
+                        },
+                    };
+
+                    // check azero balance
+                    if azero_balance > current_base_fee + ONE_AZERO  {
+                        // swap everything above base fee plus one A0
+                        let surplus = azero_balance.saturating_sub(current_base_fee).saturating_sub(ONE_AZERO);
```
deuszx, line 161: nit: why not move `let surplus = ...` above the `if` and use there instead?

relayer/relayer/src/trader/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+                        info!("{surplus} A0 above the safe limit will be swapped.");
+
+                        let path = [wrapped_azero_address.clone(), azero_ether.address.clone()];
```
deuszx, line 164: nit: path is always the same so can be moved out of the loop.

relayer/relayer/src/trader/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+                                    warn!("Error when querying for Azero ETH balance: {why:?}.");
+                                    continue;
+                                },
+                            };
+
+                        if azero_eth_balance > 0 {
+                            info!("Requesting a cross chain transfer of {azero_eth_balance} units of Azero ETH [{azero_ether_address}] to {whoami_eth}.");
+
+                            let mut receiver: [u8; 32] = [0; 32];
+                            receiver.copy_from_slice(&left_pad(eth_signed_connection.address().0.to_vec(), 32));
```
deuszx, line 226: nit: similarly to `path` - this is always the same as well so can be extracted.

relayer/relayer/src/trader/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+                        Err(why) => {
+                            warn!("Query for WETH balance failed : {why:?}.");
+                            continue;
+                        }
+                    };
+
+                    debug!("{whoami_eth} has a balance of: {wrapped_ether_balance:?} WETH9.");
+
+                    // withdraw 0xwETH -> ETH
+                    if !wrapped_ether_balance.is_zero() {
```
deuszx, line 270: Since txns on Ethereum are quite expensive - maybe it's worth setting the surplus to something higher than non-zero?

deuszx (approved):
Couldn't find anything particularly wrong with the code but it lacks one thing - tests. I understand these are complex, cross-chain tests but the trader component is important so maybe add one for happy-path to make sure there are no regressions in the future. I haven't written any here so it's not a blocker.

I'm giving my soft approval as from my POV everything looks good but I'm not that familiar with the MOST codebase.

relayer/relayer/src/trader/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+
+        info!("Starting");
+
+        select! {
+            cb_event = circuit_breaker_receiver.recv() => {
+                warn!("Exiting due to a circuit breaker event {cb_event:?}.");
+                Ok(cb_event?)
+            },
+
+            _ = async {
```
krzysztofziobro, line 136: Please use the result - in the current version all info about the error that occurred in Trader is lost and we get a mysterious `Unexpected`

relayer/relayer/src/trader/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+                            amount},
+                        Err(why) => {
+                            warn!("Query to `get_base_fee` has failed {why:?}.");
+                            continue;
+
+                        },
+                    };
+
+                    // check azero balance
+                    if azero_balance > current_base_fee + ONE_AZERO  {
```
krzysztofziobro, line 159: To perform any swaps AZERO balance should be much higher than this - we not only want to get some ETH back on the Ethereum side, but we want the 20% buffer to be enough to cover all bridging fees + txn fees + slippage.

With the current condition we would run out of the funds pretty quickly

relayer/relayer/src/trader/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+                    if let Ok (eth_balance) = eth_signed_connection.get_balance(eth_signed_connection.address(), None).await {
+                        // warning if the balance drops too low
+                        if eth_balance < ONE_ETHER.into () {
+                            warn!("{whoami_eth} has a low ETH balance: {eth_balance} ETH.");
+                        } else {
+                            info!("{whoami_eth} has a balance of {eth_balance} ETH.");
+                        }
+                    }
+
+                    sleep(Duration::from_secs(ETH_BLOCK_PROD_TIME_SEC)).await;
```
krzysztofziobro, line 293: comment: checking for it once per hour should be sufficient

krzysztofziobro (changes requested):

relayer/relayer/src/contracts/azero.rs
```diff
@@ -97,6 +262,48 @@ impl MostInstance {
...
+
+        let params = ExecCallParams::new().gas_limit(gas_limit).value(base_fee);
+
+        // Exec does dry run first, so there's no need to repeat it here
+        self.contract
+            .exec(signed_connection, "send_request", &args, params)
+            .await
+            .map_err(|why| {
+                error!("send_request failure: {why:?}");
+                AzeroContractError::SendRequestTxFailure {
```
krzysztofziobro, line 298: I think that mapping onto a generic error here is not a good idea, made sense when we didn't dry run before execution because of inscrutable errors - now revert errors should be readable

relayer/scripts/entrypoint.sh
```diff
@@ -129,5 +133,20 @@ if [[ -n "${ETH_MIN_CONFIRMATIONS}" ]]; then
   ARGS+=(--eth-tx-min-confirmations=${ETH_MIN_CONFIRMATIONS})
 fi
 
+if [[ -n "${RUN_TRADER}" ]]; then
+  ROUTER_ADDRESS=${ROUTER_ADDRESS:?"Router address is required for running trader component"}
```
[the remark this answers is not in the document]
fbielejec: take a look at #1a2dd39

relayer/relayer/src/contracts/azero.rs
```diff
@@ -39,6 +36,166 @@ pub enum AzeroContractError {
...
+        address: &str,
+        metadata_path: &str,
+        ref_time_limit: u64,
+        proof_size_limit: u64,
+    ) -> Result<Self, AzeroContractError> {
+        let address = AccountId::from_str(address)
+            .map_err(|why| AzeroContractError::NotAccountId(why.to_string()))?;
+        Ok(Self {
+            address: address.clone(),
+            contract: ContractInstance::new(address, metadata_path)?,
```
obrok, line 158: Suggestion: use `contract.address()` here too.

relayer/relayer/src/main.rs
```diff
@@ -214,6 +228,9 @@ async fn main() -> Result<(), RelayerError> {
                     tick = Instant::now();
                 }
             }
+            Err(RelayerError::TraderComponent(why)) => {
+                error!("Trader component exited with an error {why:?}. Restart is required to run Trader again.");
```
obrok, line 232: Question: so if the trader crashes we ignore it and it won't be restarted? Doesn't this mean that any connection failure in the trader will just turn it on "forever" (until relayer restart)?

relayer/relayer/src/trader/mod.rs
```diff
@@ -0,0 +1,303 @@
...
+
+use crate::{
+    config::Config,
+    connections::{azero::AzeroConnectionWithSigner, eth::SignedEthConnection},
+    contracts::{AzeroContractError, AzeroEtherInstance, MostInstance, RouterInstance},
+    helpers::left_pad,
+    CircuitBreakerEvent,
+};
+
+// trader component will sell the surplus
```
obrok, line 22: Suggestion: I'm not sure what this comment is meant to be conveying? Is it a description of this whole component? Then make it a moduledoc and perhaps expand on it. Is it a description of these two constants? Then they are inappropriately named - should be `AZERO_SELLOFF_THRESHOLD`, etc. or something like that.

```suggestion
```

relayer/relayer/src/trader/mod.rs
```diff
@@ -0,0 +1,303 @@
...
+pub const ONE_AZERO: u128 = 1_000_000_000_000;
+pub const ONE_ETHER: u128 = 1_000_000_000_000_000_000;
+
+pub const ETH_TO_AZERO_RELAYING_BUFFER: u128 = 100 * ONE_AZERO;
```
obrok, line 26: Seems to me this is the threshold, so perhaps that comment was meant to be here?

relayer/relayer/src/trader/mod.rs
```diff
@@ -0,0 +1,303 @@
...
+            azero_ether_metadata,
+            ..
+        } = &*config;
+
+        let most_azero = MostInstance::new(
+            azero_contract_address,
+            azero_contract_metadata,
+            *azero_ref_time_limit,
+            *azero_proof_size_limit,
+        )?;
```
obrok, line 87: Suggestion: This function is _huge_, so I'm looking for ways to cut it down a bit. Perhaps:

```suggestion
        let most_azero = most_instance(&config)?;
        let router = router_instance(&config)?;
        // ...
```

relayer/relayer/src/trader/mod.rs
```diff
@@ -0,0 +1,303 @@
...
+
+    #[error("Address is not an AccountId {0}")]
+    NotAccountId(String),
+
+    #[error("Trader has exited unexpectedly - this should never happen")]
+    TraderExited,
+}
+
+#[derive(Copy, Clone)]
+pub struct Trader;
```
obrok, line 60: Question: why is this a struct if the only big method does not take `&self` and it contains no data? You could just have a function instead. But see below.

relayer/relayer/src/trader/mod.rs
```diff
@@ -0,0 +1,303 @@
...
+        let wrapped_azero_address =
+            AccountId::from_str(&azero_wrapped_azero_address.clone().ok_or(
+                TraderError::MissingRequired("azero_wrapped_azero_address".to_owned()),
+            )?)
+            .map_err(|err| TraderError::NotAccountId(err.to_owned()))?;
+
+        let whoami_azero = azero_signed_connection.account_id();
+        let whoami_eth = eth_signed_connection.address().to_string();
+
+        let swap_path = [wrapped_azero_address.clone(), azero_ether.address.clone()];
```
obrok, line 121: Suggestion: Perhaps you could make the struct do some work after all and change these "helper variables" into helper functions or initialize them in `new` and then have a slimmed-down `run`?

relayer/relayer/src/trader/mod.rs
```diff
@@ -0,0 +1,303 @@
...
+                        .await {
+                            Ok(balance) => balance,
+                            Err(why) => {
+                                warn!("Error when querying for Azero ETH balance: {why:?}.");
+                                continue;
+                            },
+                        };
+
+                    // Never bridge less than 0.1 ETH
+                    if azero_eth_balance > ONE_ETHER / 10 {
```
obrok, line 224: Suggestion: If you want to comment, I'd suggest commenting about _why_ we don't want to bridge less, not just the fact that we don't.

```suggestion
                    if azero_eth_balance > BRIDGING_THRESHOLD {
```

relayer/relayer/src/trader/mod.rs
```diff
@@ -0,0 +1,303 @@
...
+                            .await
+                        {
+                            warn!("Could not send the cross-chain transfer request: {why:?}.");
+                            continue;
+                        }
+                    }
+
+                    // check ETH balance
+                    if let Ok (eth_balance) = eth_signed_connection.get_balance(eth_signed_connection.address(), None).await {
+                        // warning if the balance drops too low
```
obrok, line 249:
```suggestion
```

relayer/relayer/src/trader/mod.rs
```diff
@@ -0,0 +1,303 @@
...
+                        if eth_balance < ONE_ETHER.into () {
+                            warn!("{whoami_eth} has a low ETH balance: {eth_balance} Wei.");
+                        } else {
+                            info!("{whoami_eth} has a balance of {eth_balance} Wei.");
+                        }
+                    }
```
obrok, line 255: Question: would it be painful to extract smaller functions from, this, for example this piece, to keep the main one a manageable size? I know `async` can really make these kinds of refactors difficult, so might not be worth it...

relayer/relayer/src/trader/mod.rs
```diff
@@ -0,0 +1,303 @@
...
+            azero_signed_connection.as_connection(),
+            current_committee_id,
+            azero_signed_connection.account_id().clone(),
+        )
+        .await;
+
+    match rewards {
+        Ok(rewards) => {
+            info!("Outstanding rewards: {rewards:?}");
+            if rewards > 10 * ONE_AZERO {
```
obrok, line 290:
```suggestion
            if rewards > REWARD_WITHDRAWAL_THRESHOLD {
```

obrok (approved):
I don't see hard blockers, but I think the question of what happens when the trader crashes is worth understanding. I suspect it will be crashing relatively often, networks being unreliable and all that.

azero/artifacts/router.json
```diff
@@ -0,0 +1,2386 @@
+{
```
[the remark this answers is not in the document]
fbielejec: done

relayer/relayer/src/contracts/azero.rs
```diff
@@ -39,6 +36,166 @@ pub enum AzeroContractError {
...
```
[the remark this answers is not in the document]
fbielejec: done

relayer/relayer/src/contracts/azero.rs
fbielejec: done

relayer/relayer/src/trader/mod.rs
fbielejec: pita

krzysztofziobro (approved):

PR #185, opened 2024-05-23
Fix gas escalator usage - relayer

krzysztofziobro:


relayer/relayer/src/connections/eth.rs
```diff
@@ -22,8 +22,9 @@ use tokio::sync::Mutex;
 use crate::config::Config;
 
 pub type EthConnection = Provider<Http>;
+pub type PersistentEthConnection = GasEscalatorMiddleware<EthConnection>;
```
fbielejec, line 25: correct me if i'm wrong, bu there's nothing persistent in the connection object itself - just how you use it, so the name is a misnomer?

relayer/relayer/src/main.rs
```diff
@@ -198,8 +202,16 @@ async fn main() -> Result<(), RelayerError> {
 
     let mut tasks = JoinSet::new();
     let mut first_run = true;
+    // Gas escalator should be shared between all relayer runs
```
fbielejec, line 205: pls add why it should be shared

relayer/relayer/src/connections/eth.rs
krzysztofziobro: Yes, it's about how it's used - do you propose to remove the alias?

obrok (approved):

PR #189, opened 2024-06-11
Most AZERO & PSP22 support - Aleph Zero contracts

krzysztofziobro:
First of the PRs adding the support for PSP22 and AZERO transfers - changes in Aleph contracts:
- Possiblility to add a pair in which original token is a psp22 token from Aleph Zero chain
- `add_pair` has now an additional argument that specifies whether the original token sits on Aleph Zero chain
- A method to send native AZERO to Ethereum
- Tests


azero/contracts/most/lib.rs
```diff
@@ -898,6 +972,16 @@ pub mod most {
             Ok(())
         }
 
+        /// Set wazero(azero) psp22 token contract
+        ///
+        /// Can only be called by the contracts owner
+        #[ink(message)]
+        pub fn set_wazero(&mut self, wazero_address: AccountId) -> Result<(), MostError> {
+            self.ensure_owner()?;
```
obrok, line 980: Question: shouldn't this have `ensure_halted` too? We will most likely only ever call this once, but still?

obrok (approved):
Looks good overall, left a bunch of mostly stylistic comments.

azero/contracts/most/lib.rs
```diff
@@ -397,7 +413,7 @@ pub mod most {
         /// Upon checking basic conditions the contract will burn the `amount` number of weth tokens from the caller
         /// and emit an event which is to be picked up & acted on up by the bridge guardians.
         #[ink(message, payable)]
-        pub fn send_request_native_ether(
+        pub fn send_request_ether_to_native(
```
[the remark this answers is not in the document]
krzysztofziobro: Yes, I would be against setting a manual selector, we will need to coordinate on handling AZERO transfers anyway

azero/contracts/most/lib.rs
krzysztofziobro: Probably should

azero/contracts/most/lib.rs
```diff
@@ -409,9 +426,49 @@ pub mod most {
             // ETH_ZERO_ADDRESS as `dest_token_address` indicates native ether transfer
```
JanKuczma, line 426: This comment can be removed (or updated). Imho `NATIVE_MARKER_ADDRESS` is self-explanatory enough
```suggestion
```

JanKuczma (approved):
Looks great, good job

PR #196, opened 2024-06-25
Add architecture overview doc

obrok:


docs/most_overview.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Guardians
+
+The owner of the bridge is not a single entity, but rather a number of owners who vote on any actions with a simple majority. At the time of this writing there are 8 guardians and 5 votes are needed for any actions. Each guardian is in fact represented by 2 public addresses - a governance address and a signing address. Theoretically, the governance committee and the signing committee could have different membership (in term of business entities/persons, they _are_ different in terms of the actual keys), but that's not the case right now.
+
+In practice, a signijwng key is generated during the setup of a Guardian's infrastructure. These keys are then submitted to the governance committee. Finally, the governance committee votes to set the signing committee to the set of keys received in this manner.
```
krzysztofziobro, line 22:
```suggestion
In practice, a signing key is generated during the setup of a Guardian's infrastructure. These keys are then submitted to the governance committee. Finally, the governance committee votes to set the signing committee to the set of keys received in this manner.
```

docs/most_overview.md
```diff
@@ -0,0 +1,71 @@
...
+
+The signer is a separate component that keeps a pair of signing keys (one Ethereum and one Aleph Zero key) and communicates with the relayer over VSOCK to sign any transaction that is submitted to it with the appropriate key. This component is intended to be deployed in an AWS nitro enclave to enhance key security.
+
+### Gas Price Oracle (Aleph Zero)
+
+This contract delivers information about Ethereum gas prices to the main Most contracts on the Aleph Zero side. This is used in order to charge the appropriate amount in fees when accepting requests for Aleph Zero -> Ethereum transfers. The contract is intended to be fed with data by an external feeder, source not included in this repo. The main contract has fallback pricing setup so that it works even in the absence of the Gas Price Oracle or when it starts to provide malicious data.
+
+## Frontend
+
+The frontend for this project consists in just one tab of the overall Common frontend. As such it is contained in the [Common frontend](https://github.com/Cardinal-Cryptography/common-frontend/) repo.
```
krzysztofziobro, line 66: This is not a public repo

lesniak43 (approved):

docs/most_overview.md
```diff
@@ -0,0 +1,84 @@
...
+
+In both cases, what technically happens, is the relayer observing the chain and waiting for a `CrosschainTransferRequest` event. That event is emitted by a contract only when a valid request is processed. When such an event is observed, the relayer submits a `receiver_request` method call with the same set of data on the other chain, thereby performing its certification duty. Only members of the signing committee are allowed to call this method and the method itself enforces a majority vote before performing any transfers, so in that way only once a request is certified by a majority of the committee is it processed on the other chain.
```
krzysztofziobro, line 41:
```suggestion
In both cases, what technically happens, is the relayer observing the chain and waiting for a `CrosschainTransferRequest` event. That event is emitted by a contract only when a valid request is constructed on the source chain. When such an event is observed, the relayer submits a `receive_request` method call with the same set of data on the target chain, thereby performing its certification duty. Only members of the signing committee are allowed to call this method and the method itself requires `threshold` (signing committee parameter) of votes before performing any transfers, so in that way only once a request is certified by a predefined fraction of the committee is it processed on the other chain.
```

docs/most_overview.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Frontend
+
+The frontend for this project consists in just one tab of the overall Common frontend.
+
+## Miscellaneous
+
+### SendRequestNative
+
+As a convenience the contract on the Ethereum side provides a `SendRequestNative` method. Depending on when you're reading this, the Aleph Zero contract might have a similar functionality too. This method combines two steps - wrapping of a specified amount of Ether into Wrapped Ether and then calling a regular `SendRequest` for the same amount.
```
krzysztofziobro, line 79:
```suggestion
As a convenience the contract on the Ethereum side provides a `sendRequestNative` method. Depending on when you're reading this, the Aleph Zero contract might have a similar functionality too. This method combines two steps - wrapping of a specified amount of Ether into Wrapped Ether and then calling a regular `sendRequest` for the same amount.
```

krzysztofziobro (approved):

PR #200, opened 2024-07-14
Add evm feature

kostekIV:
Add evm feature for relayer project. 
I had to bump nightly as the cargo-contract stopped compiling with the old versions. New one introduced lints for the contract macros which cannot be solved yet by us (We cant use stable because of aleph-client dep). Also introduced changes in makefile to check/build new feature also

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -288,6 +288,11 @@ test-relayer: # Run relayer tests
 test-relayer: compile-azero-docker compile-eth
 	cd relayer && cargo test
 
+.PHONY: test-relayer-evm
+test-relayer-evm: # Run relayer tests
+test-relayer-evm: compile-azero-docker compile-eth
+	cd relayer && cargo test --features evm
+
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -307,18 +317,18 @@ solidity-lint: eth-deps
 .PHONY: relayer-lint
 relayer-lint: # Lint relayer
 relayer-lint: compile-azero-docker compile-eth
-	cd relayer && cargo clippy -- --no-deps -D warnings
+	cd relayer && cargo clippy --all-targets --all-features -- --no-deps -D warnings
 
 .PHONY: ink-lint
 ink-lint: # Lint ink contracts
 ink-lint:
-	cd azero/contracts/most && cargo clippy -- --no-deps -D warnings
```
Marcin-Radecki, line 315: Please use `-D warnings`

kostekIV: thx, done

PR #201, opened 2024-07-15
Most-90: Evm finalization changes

kostekIV:
This pr containst two important changes and few stylistic ones:
* evm finalization notion in listener - commit [b7bde16](https://github.com/Cardinal-Cryptography/most/pull/201/commits/b7bde16361ef7973fdd488acdbe777c5b2db85f6)
* evm finalizaiton notion in handler - commit [e940ee5](https://github.com/Cardinal-Cryptography/most/pull/201/commits/e940ee5e09f4a2db1c95665412ace23aa298a95a)

Commit [8bbed39](https://github.com/Cardinal-Cryptography/most/pull/201/commits/8bbed39fafed9c8bd36b6fcd67f8b05cc75da546) and larger part of commit [e940ee5](https://github.com/Cardinal-Cryptography/most/pull/201/commits/e940ee5e09f4a2db1c95665412ace23aa298a95a) contains reorganization of the code to avoid deep nested code blocks.

### DIFF
--- a/relayer/relayer/src/contracts/eth.rs
+++ b/relayer/relayer/src/contracts/eth.rs
@@ -9,10 +9,8 @@ abigen!(Most, "../eth/artifacts/contracts/Most.sol/Most.json");
 abigen!(WETH9, "../eth/artifacts/contracts/WETH9.sol/WETH9.json");
 
 pub enum SignatureState {
-    #[cfg_attr(feature = "evm", allow(dead_code))]
-    SignedNotFinalized,
+    Signed { finalized: bool },
     NeedSignature,
-    Finalized,
 }
 
 #[cfg(feature = "evm")]
@@ -28,7 +26,7 @@ pub async fn contract_signature_state<C: Middleware + 'static>(
         .block(BlockNumber::Latest)
         .await?
     {
-        Ok(Finalized)
+        Ok(Signed { finalized: true })
     } else {
         Ok(NeedSignature)
     }
@@ -47,7 +45,7 @@ pub async fn contract_signature_state<C: Middleware + 'static>(
         .block(BlockNumber::Finalized)
         .await?
     {
-        return Ok(Finalized);
+        return Ok(Signed { finalized: true });
     }
     let is_signed = !contract
         .needs_signature(request_hash, address, committee_id.into())
@@ -55,7 +53,7 @@ pub async fn contract_signature_state<C: Middleware + 'static>(
         .await?;
 
     Ok(match is_signed {
-        true => SignedNotFinalized, // Signed but not yet finalized
-        false => NeedSignature,     // Not signed and not finalized
+        true => Signed { finalized: false }, // Signed but not yet finalized
+        false => NeedSignature,              // Not signed and not finalized
     })
 }
--- a/relayer/relayer/src/contracts/eth.rs
+++ b/relayer/relayer/src/contracts/eth.rs
@@ -42,18 +42,20 @@ pub async fn contract_signature_state<C: Middleware + 'static>(
     committee_id: u128,
 ) -> Result<SignatureState, ContractError<C>> {
     use SignatureState::*;
-    let is_finalized = !contract
+    if !contract
         .needs_signature(request_hash, address, committee_id.into())
         .block(BlockNumber::Finalized)
-        .await?;
+        .await?
+    {
+        return Ok(Finalized);
+    }
     let is_signed = !contract
         .needs_signature(request_hash, address, committee_id.into())
         .block(BlockNumber::Latest)
         .await?;
 
-    Ok(match (is_signed, is_finalized) {
-        (_, true) => Finalized,          // Does not need signature on finalized block
-        (true, _) => SignedNotFinalized, // Signed but not yet finalized
-        (false, _) => NeedSignature,     // Not signed and not finalized
+    Ok(match is_signed {
+        true => SignedNotFinalized, // Signed but not yet finalized
+        false => NeedSignature,     // Not signed and not finalized
     })
 }
--- a/relayer/relayer/src/contracts/eth.rs
+++ b/relayer/relayer/src/contracts/eth.rs
@@ -1,4 +1,59 @@
-use ethers::contract::abigen;
+use aleph_client::sp_core::H160;
+use ethers::{
+    contract::{abigen, ContractError},
+    middleware::Middleware,
+    prelude::BlockNumber,
+};
 
 abigen!(Most, "../eth/artifacts/contracts/Most.sol/Most.json");
 abigen!(WETH9, "../eth/artifacts/contracts/WETH9.sol/WETH9.json");
+
+pub enum SignatureState {
+    #[cfg_attr(feature = "evm", allow(dead_code))]
+    SignedNotFinalized,
+    NeedSignature,
+    Finalized,
+}
+
+#[cfg(feature = "evm")]
+pub async fn contract_signature_state<C: Middleware + 'static>(
+    contract: &Most<C>,
+    request_hash: [u8; 32],
+    address: H160,
+    committee_id: u128,
+) -> Result<SignatureState, ContractError<C>> {
+    use SignatureState::*;
+    if !contract
+        .needs_signature(request_hash, address, committee_id.into())
+        .block(BlockNumber::Latest)
+        .await?
+    {
+        Ok(Finalized)
+    } else {
+        Ok(NeedSignature)
+    }
+}
+
+#[cfg(not(feature = "evm"))]
+pub async fn contract_signature_state<C: Middleware + 'static>(
+    contract: &Most<C>,
+    request_hash: [u8; 32],
+    address: H160,
+    committee_id: u128,
+) -> Result<SignatureState, ContractError<C>> {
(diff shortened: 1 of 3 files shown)


relayer/relayer/src/listeners/eth.rs
```diff
@@ -154,26 +154,44 @@ impl EthereumListener {
     }
 }
 
+#[cfg(feature = "evm")]
```
DamianStraszak, line 160: Ethereum is also evm :) Maybe call it differently? layer2, l2, rollup, arbitrum?

Marcin-Radecki: `relayer-eth` and `relayer-l2` works for me (names of artifacts)
so cargo feature name would be just `l2` in such case

kostekIV: ok, but lets do it not here. I have created https://cardinal-cryptography.atlassian.net/browse/MOST-95

relayer/relayer/src/handlers/azero.rs
```diff
@@ -170,9 +173,7 @@ impl AlephZeroEventHandler {
                     // Dry-run the tx to check for potential reverts.
                     call.clone().gas(config.eth_gas_limit).call().await?;
 
-                    info!(
-                        "Sending tx for request 0x{request_hash_hex} to the Ethereum network and waiting for {eth_tx_min_confirmations} confirmations."
-                    );
+                    info!("Sending tx for request 0x{request_hash_hex} to the Ethereum network and waiting for {eth_tx_min_confirmations} confirmations.");
```
ggawryal, line 176: Maybe we can rename all `Ethereum` in logs to `EVM chain` to have the same name for both bridges?

relayer/relayer/src/listeners/eth.rs
```diff
@@ -66,120 +66,137 @@ impl EthereumListener {
...
+                Ok(unprocessed_block_number) = next_unprocessed_block_number.recv() => {
+                    unprocessed_block_number
+                }
+            };
 
-                Ok (unprocessed_block_number) = next_unprocessed_block_number.recv() => {
-                    // Query for the next unknown finalized block number, if not present we wait for it.
-                    info!(target: "EthereumListener","Waiting for the next finalized block number");
+            // Query for the next unknown finalized block number, if not present we wait for it.
+            info!(target: "EthereumListener", "Waiting for the next finalized block number");
```
ggawryal, line 80: Since you're already there, can we extract log target? :P

relayer/relayer/src/contracts/eth.rs
```diff
@@ -1,4 +1,61 @@
...
+    middleware::Middleware,
+    prelude::BlockNumber,
+};
 
 abigen!(Most, "../eth/artifacts/contracts/Most.sol/Most.json");
 abigen!(WETH9, "../eth/artifacts/contracts/WETH9.sol/WETH9.json");
+
+pub enum SignatureState {
+    #[cfg_attr(feature = "evm", allow(dead_code))]
+    SignedNotFinalized,
```
ggawryal, line 13: Alternatively, we could disable this variant totally when `evm` feature is enabled: this could also help us find all the places where this variant is used, but I imagine this could make test code more complicated in some cases. wdyt?

ggawryal (approved):
Nice!

relayer/relayer/src/handlers/azero.rs
```diff
@@ -71,90 +74,90 @@ impl AlephZeroEventHandler {
...
-                        return Ok(());
-                    }
-                }
+        match &event.name {
+            Some(name) if name.eq("CrosschainTransferRequest") => {}
+            _ => {
+                debug!("Skipping non azero contract event");
+                return Ok(());
+            }
+        }
```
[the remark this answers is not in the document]
kostekIV: Did something similar to the `if event ...`

relayer/relayer/src/listeners/eth.rs
kostekIV: Done

relayer/relayer/src/contracts/eth.rs
kostekIV: I have refactored it a bit to solve this issue

relayer/relayer/src/handlers/azero.rs
kostekIV: I kinda dont want to do it, at least not yet :P

relayer/relayer/src/listeners/eth.rs
Marcin-Radecki: Yes, log targets are misssing in this repo in general, so does public API rustdocs. We should improve them as we go.

relayer/relayer/src/handlers/azero.rs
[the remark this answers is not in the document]
ggawryal: Lovely

relayer/relayer/src/contracts/eth.rs
ggawryal: Not exactly what I meant: this still doesn't prevent us from accidentally using `Signed {finalized = false }` with `evm` feature, but it's not that important

relayer/relayer/src/handlers/azero.rs
```diff
@@ -71,90 +74,90 @@ impl AlephZeroEventHandler {
...
-                        warn!("Skipping blacklisted request: 0x{request_hash_hex}");
-                        return Ok(());
-                    }
-                }
+        if !event
+            .name
+            .is_some_and(|name| name.eq("CrosschainTransferRequest"))
+        {
+            debug!("Skipping non azero contract event");
+            return Ok(());
```
Marcin-Radecki, line 82: Kudos for an effort to make this big if body shorter and more readable

relayer/relayer/src/handlers/azero.rs
```diff
@@ -71,90 +74,90 @@ impl AlephZeroEventHandler {
...
+        }
+
+        let data = event.data;
+
+        // decode event data
+        let crosschain_transfer_event @ CrosschainTransferRequestData {
```
Marcin-Radecki, line 88: Nitpick: this `@ sth` syntax seems to me like a seldom used Rust feature (https://doc.rust-lang.org/book/ch18-03-pattern-syntax.html#-bindings), and mostly in `match` patterns. It takes a minute for a reader to Google what it does. `get_request_event_data` is used only once and here, so we can move this `debug!` inside so we don't need binding syntax and just unpack `CrosschainTransferRequestData` `struct` as usual

relayer/relayer/src/listeners/eth.rs
```diff
@@ -61,125 +62,142 @@ impl EthereumListener {
...
     }
 }
 
-pub async fn get_next_finalized_block_number_eth(
+#[cfg(feature = "evm")]
+pub async fn get_next_finalized_block_number(
+    eth_connection: Arc<EthConnection>,
+    not_older_than: u32,
+) -> u32 {
+    // In evm context we treat latest block as finalized.
```
Marcin-Radecki, line 166: Why it is so? Which stage are we (https://docs.arbitrum.io/how-arbitrum-works/tx-lifecycle) ?
To ask this question differently: are we fine with 'soft finality' here (https://docs.arbitrum.io/for-devs/troubleshooting-building#how-many-block-numbers-must-we-wait-for-in-arbitrum-before-we-can-confidently-state-that-the-transaction-has-reached-finality)

Marcin-Radecki (approved):

relayer/relayer/src/contracts/eth.rs
```diff
@@ -1,4 +1,59 @@
...
+    address: H160,
+    committee_id: u128,
+) -> Result<SignatureState, ContractError<C>> {
+    use SignatureState::*;
+    if !contract
+        .needs_signature(request_hash, address, committee_id.into())
+        .block(BlockNumber::Latest)
+        .await?
+    {
+        Ok(Signed { finalized: true })
```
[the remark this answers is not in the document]
kostekIV: But these are internal of the contract call. We would need to call it anyway to check the current committee and wheter request_hash is not in processedRequests etc. We could maintain some cache with that to save 1 call in such situations but I dont htink its worth it right now

relayer/relayer/src/listeners/eth.rs
kostekIV: Yep, we are fine with it :p more details in the channel dedicated to this topic

relayer/relayer/src/contracts/eth.rs
```diff
@@ -1,4 +1,59 @@
...
 
 abigen!(Most, "../eth/artifacts/contracts/Most.sol/Most.json");
 abigen!(WETH9, "../eth/artifacts/contracts/WETH9.sol/WETH9.json");
+
+pub enum SignatureState {
+    Signed { finalized: bool },
+    NeedSignature,
+}
+
+#[cfg(feature = "evm")]
```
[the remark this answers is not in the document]
kostekIV: will do

relayer/relayer/src/handlers/azero.rs
kostekIV: done

PR #202, opened 2024-07-17
Fixes plus include scripts in ci

kostekIV:
Some adjustments to the scripts to make them work. Also added them to ci that runs on merge to main since they are rather long.

You can check here ci pass: https://github.com/Cardinal-Cryptography/most/actions/runs/9979956139/job/27580250622

### DIFF
--- a/.github/workflows/test-transfers.yml
+++ b/.github/workflows/test-transfers.yml
@@ -26,7 +26,6 @@ jobs:
         timeout-minutes: 30
         shell: bash
         run: |
-          exit 1
           make local-bridgenet
           sleep 60 # wait 1 minute for everything to be running
           make deploy-docker
--- a/.github/workflows/test-transfers.yml
+++ b/.github/workflows/test-transfers.yml
@@ -3,9 +3,6 @@
 name: Transfer test
 
 on:
-  pull_request:
-    branches:
-      - master
   push:
     branches:
       - master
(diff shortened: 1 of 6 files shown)


.github/workflows/test-transfers.yml
```diff
@@ -0,0 +1,33 @@
+# Runs bridge and perform transfers from and to ethereum.
```
Marcin-Radecki, line 1: Please add slack notifications as per https://github.com/Cardinal-Cryptography/aleph-node/blob/main/.github/workflows/nightly-fe-e2e-tests.yml#L151-L164 , please use `SLACK_INCOMING_WEBHOOK_MOST_CI_NOTIFICATIONS` secret

Marcin-Radecki (approved):

.github/workflows/test-transfers.yml
```diff
@@ -0,0 +1,33 @@
...
+        shell: bash
+        run: |
+          make local-bridgenet
+          sleep 60 # wait 1 minute for everything to be running
+          make deploy-docker
+          DOCKER_RELAYER_COMPILE_CONTRACTS=no-compile make run-relayers
+          
+          make ethereum-azero-transfer
+          sleep 300 # wait 5 minutes for transfer to register on both chains
+          make azero-ethereum-transfer
```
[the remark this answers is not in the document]
kostekIV: This test is mostly for checking whether the scripts we use in the repo works as in compiles/runs correctly. THe azero-transfer should fail tho if the eth transfer did not succeed

.github/workflows/test-transfers.yml
kostekIV: done

Marcin-Radecki (approved):
Splendid!

PR #203, opened 2024-07-18
Most-95: Rename evm -> l2

kostekIV:


relayer/relayer/src/listeners/eth.rs
```diff
@@ -158,21 +158,21 @@ impl EthereumListener {
     }
 }
 
-#[cfg(feature = "evm")]
+#[cfg(feature = "l2")]
 pub async fn get_next_finalized_block_number(
```
Marcin-Radecki, line 162: Can this method be moved to `l2_eth` module, and exported? Motivation is to minimize `#[cfg(...)]` annotations

Marcin-Radecki (approved):

relayer/relayer/src/listeners/eth.rs
kostekIV: After considerations I moved this closer to the EthConnection struct to eth.rs file.I feel like l2_eth file would not help with ot that much and for now this is good enough and we can refactor it properly if we have time later, wdyt?

Marcin-Radecki: Legit, thanks

PR #204, opened 2024-07-19
Remove trader

maciejzelaszczyk:
This removes the Trader component. An attempt at a call setting the payout account is made. What remains to be done is to: polish this call, add the possibility of passing this account via an env variable and add a script for actually paying the rewards out (from the point of view of the relayer).


relayer/relayer/src/main.rs
```diff
@@ -274,6 +269,25 @@ async fn run_relayer(
...
+
+    let current_committee_id = most_azero
+        .current_committee_id(azero_connection.as_connection())
+        .await?;
+    most_azero
+        .set_payout_account(
+            &azero_signed_connection,
+            current_committee_id,
+            AccountId::from_string(&config.payout_address)
+                .map_err(|why| AzeroContractError::NotAccountId(why.to_string()))?,
```
Marcin-Radecki, line 287: This should be backward compatible - let's return `None` if there's no  such entry `config`

relayer/scripts/entrypoint.sh
```diff
@@ -143,33 +139,18 @@ if [[ -n "${ETH_MIN_CONFIRMATIONS}" ]]; then
...
-
 if [[ "${ETH_GAS_LIMIT}" =~ ^[0-9]+$ ]]; then
   echo "Setting --eth-gas-limit to ${ETH_GAS_LIMIT}"
   ARGS+=(--eth-gas-limit=${ETH_GAS_LIMIT})
 fi
 
+if [[ -n "${PAYOUT_ADDRESS}" ]]; then
+  ARGS+=(--payout-address=${PAYOUT_ADDRESS})
+else
+  echo "Payout address required but not provided." 1>&2
```
Marcin-Radecki, line 150: Please make this change backward compatible

relayer/relayer/src/main.rs
```diff
@@ -274,6 +269,25 @@ async fn run_relayer(
...
+        )
+        .await?;
```
kostekIV, line 289: Pls extract to function

relayer/relayer/src/main.rs
maciejzelaszczyk: I have changed this such that the parameter is now optional.

relayer/relayer/src/main.rs
maciejzelaszczyk: Extracted.

kostekIV (approved):
lgtm

relayer/scripts/run_relayer.sh
```diff
@@ -42,6 +42,13 @@ if [[ -n "${SIGNER_CID}" ]]; then
   ARGS+=(--signer-cid=${SIGNER_CID})
 fi
 
+if [[ -n "${PAYOUT_ADDRESS}" ]]; then
+  ARGS+=(--payout-address=${PAYOUT_ADDRESS})
+else
+  echo "Payout address required but not provided." 1>&2
+  exit 1
```
Marcin-Radecki, line 49: Why do we treat lack of `PAYOUT_ADDRESS` as error condition here?

Marcin-Radecki (approved):

relayer/scripts/run_relayer.sh
maciejzelaszczyk: We shouldn't be doing this - I missed it. Fixed now.

PR #213, opened 2024-08-01
MOST-119: Most L2 in ink

kostekIV:
Unfortunately cp pasting of the origanl l1 most with a lot of code removed. A lot of functions related to fee, pocket money, rewards have been removed. Code related to minting/burnign also removed. Checking of locality of the token also removed.

If we will have time we can tidy this up by extracting common functionality to the shared library (and I would love it)


obrok (approved):
Looks good. As this is a copy of the existing contract with some parts removed, I mostly looked at the diff between those 2.

Marcin-Radecki (approved):
Good job!

PR #222, opened 2024-08-08
Most-133: azero scripts

kostekIV:

PR #227, opened 2024-08-10
Use NATIVE_ADDRESS_MARKER for senRequestNative

kostekIV:


eth/test/MostL2.js
```diff
@@ -180,9 +166,7 @@ describe("MostL2", function () {
     });
 
     it("Emits correct event", async () => {
-      const { most, bazeroAdrressBytes } = await loadFixture(
-        deployEightGuardianMostFixture,
-      );
+      const { most } = await loadFixture(deployEightGuardianMostFixture);
```
Marcin-Radecki, line 169: in `deployEightGuardianMostFixture`, we still call `await most.addPair(bazeroAdrressBytes, bazeroAdrressBytes, false);`, but we don't do it anymore in `eth/scripts/1_l2_setup_bridge_contracts.js`. So not sure which version is correct

eth/contracts/MostL2.sol
```diff
@@ -166,7 +168,7 @@ contract MostL2 is AbstractMost {
         );
 
         // transfer native
-        if (_destTokenAddress == bAzeroAddress) {
+        if (_destTokenAddress == NATIVE_MARKER_ADDRESS) {
```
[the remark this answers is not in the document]
kostekIV: Good catch, I forgot to rm this require.

eth/test/MostL2.js
kostekIV: The 1_l2_* is correct. Fixed

Marcin-Radecki: Do we need `bazeroAdrressBytes` at all returned by `deployEightGuardianMostFixture`? I mean in those tests, we call `receiveRequest` always with `bazeroAdrressBytes` as `destTokenAddress` argument. I mean, those tests pass right, but do they test native transfers?

kostekIV: Good catch again. Sorry I was doing it too fast :P. Anyway, fixed the tests (see the usage of abs)

azero/contracts/most-l2/lib.rs
```diff
@@ -339,14 +338,9 @@ pub mod most_l2 {
                 .transferred_value(amount_to_bridge)
                 .invoke()?;
 
-            let dest_token_address = self
-                .supported_pairs
```
Marcin-Radecki, line 343: Do we want to remove this logic from contracts on AZERO side? https://github.com/Cardinal-Cryptography/most/pull/222/files#diff-ed93ab43297ff1acce2388424f1b571f305e5686aba3463a5d27d1b0393040eaR128

Marcin-Radecki (approved):

azero/contracts/most-l2/lib.rs
kostekIV: This is removed here

PR #229, opened 2024-08-12
Fix native transfer

kostekIV:


azero/contracts/most-l2/lib.rs
```diff
@@ -252,7 +252,7 @@ pub mod most_l2 {
             dest_receiver_address: [u8; 32],
             native_azero_request: bool,
         ) -> Result<(), MostError> {
-            if dest_receiver_address == ZERO_ADDRESS {
+            if dest_receiver_address == ZERO_ADDRESS && !native_azero_request {
```
Marcin-Radecki, line 255: I'm not sure how this fix works - for `send_request_native_azero`, the `dest_token_address` is 0, not `dest_receiver_address`. Or do you mean that for `send_request_native_azero`,  the `dest_receiver_address` must be 0?

PR #232, opened 2024-08-14
When dev env is set up, we should lock wAZERO

Marcin-Radecki:


azero/scripts/2_l2_setup.ts
```diff
@@ -39,6 +40,18 @@ async function main(): Promise<void> {
   console.log(`wAZERO address set to ${wAzeroAddress}`);
   await most.tx.setWazero(wAzeroAddress);
 
+  if (config.dev) {
+    console.log('Creating WrappedAzero instance');
+    const wAzero = new WrappedAzero(wAzeroAddress, deployer, api)
+    const deposit = BigInt(10000 * 10 ** 12);
```
kostekIV, line 46: we should get this value from the config

PR #246, opened 2024-09-24
Most-185: Decimal conversion

kostekIV:


azero/contracts/most/lib.rs
```diff
@@ -542,6 +547,7 @@ pub mod most {
                     .contains::<AccountId>(dest_token_address.into());
 
                 if dest_token_address == NATIVE_MARKER_ADDRESS {
+                    let amount = amount.saturating_div(RATIO);
```
Marcin-Radecki, line 550: To be safe, let's add a check `amount >= 10e6` in `sendRequestAzeroToNative` in `AbstractMost.sol`. We have a similar check in MOST L2. This check is because we don't want to end up in a situation in which the `amount` here is 0.

Marcin-Radecki (approved):

azero/contracts/most/lib.rs
Marcin-Radecki: This will be done in another PR.

PR #247, opened 2024-09-25
MOST-174: Min transfer value

obrok:


eth/contracts/AbstractMost.sol
```diff
@@ -381,6 +397,12 @@ abstract contract AbstractMost is
         wrappedAzeroAddress = _wrappedAzeroAddress;
     }
 
+    function setTransferLimit(
+        ITransferLimit _transferLimit
+    ) external onlyOwner {
```
Marcin-Radecki, line 402: Should it have `whenPaused` here?

eth/hardhat.config.js
```diff
@@ -119,15 +120,21 @@ var config = {
   },
 };
 
+var SEPOLIA_ACCOUNT;
```
Marcin-Radecki, line 123: Nice, I added this in parallel as well to https://github.com/Cardinal-Cryptography/most/pull/248, withou seeing this PR first :)

eth/contracts/AbstractMost.sol
obrok: I don't think it's needed, adding/removing/changing won't invalid any requests already in the queue.

eth/hardhat.config.js
obrok: :+1: 
It's going to be fun to merge now :)

eth/contracts/TransferLimit.sol
```diff
@@ -0,0 +1,87 @@
...
+
+        USDMinimum memory config = usdMinima[_token];
+        if (config.limit > 0) {
+            (, int usdPrice, , , ) = config.oracle.latestRoundData();
+            uint256 usdBasedMinimum = config.limit / uint256(usdPrice);
+
+            minimum = Math.min(minimum, usdBasedMinimum);
+        }
+
+        return minimum;
```
youPickItUp, line 75: I wonder if we could round it up a little. Now the minimal value might (and in case of USD limit probably will) look like  `1.169847813119247738`. It's a peculiar requirement that I have to display on front-end :sweat_smile:

eth/contracts/AbstractMost.sol
```diff
@@ -40,6 +41,8 @@ abstract contract AbstractMost is
 
     address public wrappedAzeroAddress;
 
+    ITransferLimit public transferLimit;
```
kostekIV, line 44: Doesnt this makes l2 and l1 most unupgradable?

kostekIV: Didnt we want limits only for l1 most since we have fees in both direction on the l2?

Marcin-Radecki: Agree with the second comment. About the first one, I've just tried upgrading to L1, and it worked. Not sure if this line breaks L2 upgrade, have not checked.

kostekIV: You are right l1 is still fine but l2 should break since it have fields in the mostl2.sol

Marcin-Radecki: Let's then remove most2.sol entry and we should be good

kostekIV: we are a littlebit to late with https://eips.ethereum.org/EIPS/eip-7201 since l1 most was already deployed

kostekIV: > Let's then remove most2.sol entry and we should be good 
We would need to remove also new field from abstractmost and move it to l1most

Marcin-Radecki (approved):

eth/contracts/TransferLimit.sol
```diff
@@ -0,0 +1,87 @@
...
+    /// @param _token The token address
+    function minimumTransferAmount(
+        address _token
+    ) public view returns (uint256) {
+        uint256 minimum = defaultMinima[_token];
+
+        USDMinimum memory config = usdMinima[_token];
+        if (config.limit > 0) {
+            (, int usdPrice, , , ) = config.oracle.latestRoundData();
+            uint256 usdBasedMinimum = config.limit / uint256(usdPrice);
```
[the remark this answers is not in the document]
obrok: Hm, good point... I guess it theoretically could be. But as the code stands it should revert, so in MOST it will just ignore `TransferLimit` in that case - I think that's OK.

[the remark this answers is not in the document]
kostekIV: Division by zero is defined in evm tho. It returns 0 in such case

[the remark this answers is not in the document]
obrok: https://consensys.github.io/smart-contract-best-practices/development-recommendations/deprecated/division-by-zero/ This says that for solidity > 0.4 it should revert?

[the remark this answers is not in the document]
kostekIV: You are right, mb, @Marcin-Radecki from my side we can resolve the conversation

eth/contracts/AbstractMost.sol
obrok: Done - @Marcin-Radecki want to try the upgrade again?

Marcin-Radecki (approved):

eth/contracts/AbstractMost.sol
Marcin-Radecki: Yes, it works.

PR #254, opened 2024-10-02
Fixing almost all e2e cases. Added Mainnet deployment configuration.

Marcin-Radecki:
This PR brings back unpausing MOST on both sides during dev, which was erroneously removed in the other PR. Also, there was some problem with particular method ETH most is unpaused, hence I used the same method which MOST L2 is unpaused.

This fixes all e2e tests except Eth -> L1 AZERO one.

### DIFF
--- a/e2e-tests/src/balance.rs
+++ b/e2e-tests/src/balance.rs
@@ -91,7 +91,7 @@ impl Balance {
     /// Bridge AZERO to wAZERO.
     pub fn bridge_azero_azero_to_eth(&self, transfer_amount: u128) -> Result<Self> {
         let mut balance = self.clone();
-        let azero_wzaero_decimal_ratio = 1000000u128;
+        let azero_wzaero_decimal_ratio = 1_000_000;
         balance.azero = balance
             .azero
             .checked_sub(transfer_amount)
--- a/e2e-tests/src/balance.rs
+++ b/e2e-tests/src/balance.rs
@@ -91,13 +91,14 @@ impl Balance {
     /// Bridge AZERO to wAZERO.
     pub fn bridge_azero_azero_to_eth(&self, transfer_amount: u128) -> Result<Self> {
         let mut balance = self.clone();
+        let azero_wzaero_decimal_ratio = 1000000u128;
         balance.azero = balance
             .azero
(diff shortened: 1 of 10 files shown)


e2e-tests/src/balance.rs
```diff
@@ -91,13 +91,14 @@ impl Balance {
     /// Bridge AZERO to wAZERO.
     pub fn bridge_azero_azero_to_eth(&self, transfer_amount: u128) -> Result<Self> {
         let mut balance = self.clone();
+        let azero_wzaero_decimal_ratio = 1000000u128;
```
kostekIV, line 94:
```suggestion
        let azero_wzaero_decimal_ratio = 1_000_000u128;
```

kostekIV: is u128 needed?

eth/scripts/1_setup_bridge_contracts.js
```diff
@@ -80,7 +73,7 @@ async function main() {
   // --- setup
   const Most = await ethers.getContractFactory("Most");
   const most = Most.attach(addresses.most);
-
+  //
```
kostekIV, line 76: nit: rm

eth/scripts/1_setup_bridge_contracts.js
```diff
@@ -92,17 +85,6 @@ async function main() {
     );
   }
 
-  // --- add Aleph -> Ethereum token pairs
```
kostekIV, line 95: why this change?

kostekIV (approved):

eth/scripts/1_setup_bridge_contracts.js
Marcin-Radecki: We removed this pair when we merged change https://github.com/Cardinal-Cryptography/most/commit/f2bedb59fd98f884ddd4909c29484122508ba81e, so that one must brigde native AZERO with `sendRequestAzeroToNative`

eth/scripts/1_setup_bridge_contracts.js
Marcin-Radecki: done

